PR #21, opened 2018-03-20
Add SGF file load/save

cngoodboy:
* Add move number show
* Add SGF file load/save


CamWagner:
wow you added a lot! I'll review these as soon as possible :)

PR #45, opened 2018-03-31
Bug fix: Open SGF error when SGF file have unicode charactors

cngoodboy:
I refactored the code to load the SGF file part


featurecat:
thanks :)

Issue #52, opened 2018-04-01
Highlight next move instead of top move


dfannius:
I am using Lizzie primarily to analyze SGF files of games that have already been played. It would be much more useful to me if Lizzie highlighted the location of the move that was actually played in the displayed position (if any) rather than the location of its top-ranked move (which you can already see on the board from the color and stats of the suggestions). This could of course be an option if people prefer the old behavior.

dfannius:
Implemented with #53 (and I found a way to call out both the next move and Leela's top move).

PR #53, opened 2018-04-01
Highlight next move

dfannius:
Implements the feature request of #52. The next move of the game (if any) is circled in black or white, depending on the player, slightly outside the red circle of Leela Zero's top move so that they are visible independently. See if you like it.


featurecat:
Thanks :)

PR #54, opened 2018-04-02
Add plug-in feature for extending Lizzie

icespike-x:
This allows some of the less frequently used features not to be loaded, but also allows the user to add custom functionality without modifying Lizzie.jar.


featurecat:
could you please explain more about how. plug-ins work? looks exciting.

cngoodboy:
Runtime load jar file and invoke methods

cngoodboy:
@featurecat I will write a document about how to development a plug-in

PR #57, opened 2018-04-03
fixes NPE.

bvandenbon:
Yesterday I launched the application on my home laptop and everything worked fine.
During lunch break I tried the same at work but had a NullPointerException.
So, I fixed the NPE, and tested, it works fine now.

This is actually my first github commit ever.
So I do hope I followed procedure correctly.


featurecat:
looks good :) thanks.

PR #58, opened 2018-04-03
Don't react to parentheses inside tags

dfannius:
Fixes a bug where parentheses inside tags (notably comments) would cause the SGFParser to exit early.


featurecat:
thanks :)

Issue #32, opened 2018-03-28
Features for Game Record Analysis


stephenmartindale:
Lizzie 0.3 is great. I am super impressed. But my absolute top No. 1 use case is the analysis of games loaded from SGF files and I think that a few very minor improvements could make Lizzie 10 times better for that purpose.

When you are looking at a node which there ARE existing children in the SGF file, you currently have to press next and then back so that you know what the original players played or considered because Lizzie only draws Leela's ideas. This is sub-optimal but sometimes sufficient but, if the next move(s) are not part of Leela's top suggestions, you get no information about it/them. So...

1. It would be cool if you give an indication not only of what Leela is thinking about but also the existing child nodes in the SGF and, if one of Leela's top moves *is* an existing child, mark it in some way.
2. It would be nice if you include the existing children and Leela's valuation of those moves if they existing in the SGF even if Leela doesn't consider them to be good or worth considering.

Finally, there's little reason to analyze a game that's over so, by default, loaded SGF files should start at move 1, not at the end.

On a side note, the readme/docs could use some explanation of the numbers and colouring scheme for Leela's moves.

Carry on. This is looking grand.

featurecat:
good suggestions. what is the difference between suggestions 1 and 2? they seem like they're suggesting the same thing.

stephenmartindale:
In essence, they are the same thing. Except that it might be needed to force leela to consider a certain child node in some way -- perhaps there is some way to make leela spend a minimum of 2% of its evaluation effort on each pre-existing child, instead of simply neglecting to expand those branches because they are such low-utility subtrees? I don't know if this is directly possible or not but it could always be hacked by moving one node down the branch and asking Leela to evaluate *that* for a certain fraction of the time.

ProLeela:
First of all thank you for the nice job.
Three suggestions, maybe useful (some of them maybe claimed also by others):

1. A GUI with menus and buttons.

2. Possibility to navigate in variants. Maybe the user wants to test a move or an entire variant then to came back in the main variant and continue with the AI analysis... And also maybe he wants to save in the sgf all this explored variants, beside the main analysis.

The idea being that in the final he will obtain a sgf with all the explored variants (of course if the scores can also be included this would be ideal :D)

3 (I am not sure if this is possible, probably depends directly of Leela Zero) Maybe I have a move not in the list of AI's moves/spots on the board. I want to know the score AI gives to my move.

dfannius:
[...]

PR #59, opened 2018-04-03
Start at beginning of game after loading SGF file

dfannius:
Implements one request from issue #32, which I wanted as well: when an SGF file is loaded, we start at the beginning of the game, not the end.


cngoodboy:
I think it 's make no sense

PR #60, opened 2018-04-03
Remember variation tree, easily return to main trunk

dfannius:
This partially implements @ProLeela's request in issue #32.

From a user standpoint, Lizzie now remembers the entire variation tree that has been explored. Right-arrow always chooses the original move played in the current position (this is the move from the SGF file if it exists), and that is the move displayed with a black or white outline. This means that you can now explore a variation, left-arrow back to the main trunk, and then continue with the loaded game (previously, the rest of the loaded game would be blown away).

From a coder standpoint, this code should probably be examined a little more closely than my other pull requests. I'm not a Java programmer and don't know whether my use of `ArrayList`s to store child `BoardHistoryNode`s is inefficient with respect to speed or memory. I don't do any cleanup and am assuming that Java's GC just takes care of freeing everything as necessary.

### DIFF
--- a/src/wagner/stephanie/lizzie/rules/Board.java
+++ b/src/wagner/stephanie/lizzie/rules/Board.java
@@ -101,7 +101,7 @@ public class Board {
                 Lizzie.leelaz.sendCommand("genmove " + (history.isBlacksTurn()? "W" : "B"));
 
             // update history with pass
-            history.add(newState);
+            history.addOrGoto(newState);
 
             Lizzie.frame.repaint();
         }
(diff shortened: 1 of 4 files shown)


featurecat:
great! thank you. the code looks good and follows the conventions well :). Java does automatically gc objects with no references so it should be fine with memory. 

since the max number of next nodes is ~361, I'm not really concerned with speed. arraylist is good because in the future we may consider the index of the last accessed node when using redo, like in Sabaki / KGS.

PR #66, opened 2018-04-04
Handicap load/save support

cngoodboy:


cngoodboy:
Sorry, I 'm not check it

PR #67, opened 2018-04-04
Restructured source code and improved maven build script

aerisnju:
I find that Lizzie has adopted apache maven as a 3rd parth library management and build system. But according to maven recommendations, the current project directory layout is not comforming to the [Standard Directory Layout](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html), which makes the original pom.xml unable to build. Later @cngoodboy had made some changes to make the script worked. But this is still not a good practice for using maven. To improve it, I made several changes listed below:

- Moved all source code packages to `src/main/java` according to the [Standard Directory Layout](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) and moved all fonts to `src/main/resources`, which are both described in the maven document. The fonts in resources directory will automatically be packaged in the built jar file.
- Improved `pom.xml` to auto package a standalone jar file, with all of its dependencies automatically packaged into it. (Currently the only dependency is the json library) The built jar file can run directly without any dependencies. The produced file is located in the `target/` directory with the name `lizzie-<version>-shaded.jar`.(This can be changed by specifying the output file name in `pom.xml`)
- Removed `lib/` directory with the json library file. This is because maven automatically stores its libraries in a local repo, which means there is no need to keep it in the source code repo. 

I did't change any code during the refactoring except `pom.xml`. I know this is a big change to the source code repo, but I hope the change can make the build script simpler (some tricky configurations can be eliminated since we follow the convention), and the other contributors may feel more familiar if they have some maven experiences. Wish for your careful review and thank you for this amazing project!


featurecat:
Thanks for your contribution. I haven't worked with a maven project before, this is probably more familiar to maven users :)

PR #68, opened 2018-04-04
add support for pass moves.

bvandenbon:
I encountered a NullPointerException when I tried to save an SGF file.
The reason was that LeeLa played a pass move.

Pass moves should be in format ";B[tt]" according to the SGF standard.

I changed a couple of things to this method:
1) a try-finally mechanism assures that the file is always properly closed.
2) if no color is found, there is no longer a "return" statement, but a "continue" statement.
3) support for the pass moves.
4) some cosmetics: line-breaks/comments/eliminated a StringBuilder.

Everything is tested and works great :).

### DIFF
--- a/src/wagner/stephanie/lizzie/rules/SGFParser.java
+++ b/src/wagner/stephanie/lizzie/rules/SGFParser.java
@@ -130,29 +130,42 @@ public class SGFParser {
     public static boolean save(String filename) throws IOException {
         FileOutputStream fp = new FileOutputStream(filename);
         OutputStreamWriter writer = new OutputStreamWriter(fp);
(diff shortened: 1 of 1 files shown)


featurecat:
it looks good to me! thanks

featurecat:
One of the next features we need for this is loading/saving handicap game stones, maybe by preserving all header info from the original loaded sgf. if you want to write that, I'd appreciate it :D

bvandenbon:
Here you go.
Just made a pull request.
I hope that's what you had in mind.

2 additional shortcuts:

   - n = new game dialog (which allows setting names of players, komi and
   handicap.)
   - i = edit info dialog (which can be used at all times, but does not
   allow editing handicap or komi for an active game.)

kind regards,
Bram Vandenbon

PR #69, opened 2018-04-04
New Game Dialog and Edit Info Dialog + support for handicap stones + game info is saved to SGF.

bvandenbon:
New Game dialog (shortcut n) which allows setting the player names, komi and handicap. The player names can be editted afterwards using the edit info dialog (shortcut i). The handicap settings are also communicated to LeeLa Zero, and finally it is possible to save everything to SGF. (including also the date of the game).


isty2e:
Is it any meaningful to set the komi value manually? LZ only accepts 7.5, so I do not think this is a valid option after all...

cngoodboy:
I have a simpler plan

bvandenbon:
Agreed, some things are not as I expected them.
This seems to be a difference between the LeeLa and the LeeLa Zero implementation.
Because LeeLa actually does support it.

On the other hand, you can actually set the komi using a GTP command. It's just that it does not calculate the values of the moves correctly. (e.g. Winning percentage is still 50-50 if you play with a komi of 100).

But if we can't set komi, I do find it strange to play a handicap game with a 7.5 komi. 
LeeLa should be using 0.5 komi in those cases.
(I've been looking for a workaround, e.g. modifying the number of captured stones as an invisible alternative to komi. But at first sight, that's impossible in GTP).

Can you give me some feedback about the other features as well ?
- the New Game Dialog and Game Info Dialogs.
- Saving the player names and date to SGF.

isty2e:
The number of captured stones does not affect LZ at all; It is not even in the input and is not used for scoring. There is no workaround here, unfortunately.

bvandenbon:
Perhaps I should rework it first. Taking your feedback in account.

featurecat:
sounds good

featurecat:
did you rework it since then?

bvandenbon:
We probably crossed each other.
I just made a new pull request (a different one) with a lot of changes.

2018-04-14 15:54 GMT+02:00 featurecat <notifications@github.com>:

PR #71, opened 2018-04-05
Add handicap support

cngoodboy:


featurecat:
Looks like this was already resolved by #91.

PR #72, opened 2018-04-05
Widen the margin of board so that the stones will not hide coordinates

aerisnju:
Lizzie now can show the coordinates to help place or record moves by pressing the shortcut "C". But I find that if a stone was placed on the edge of the board(that is, Row 1 or 19, or Column A or T), it would hide the coordinate texts, because the coordinate texts are too close to the board lines.
I managed to find a simple and quick solution for this: just increased the value of `BoardRenderer.MARGIN` from `0.03` to `0.06`, so that there would be enough room between the coordinate texts and the stones and the stones would not hide the texts no longer.


featurecat:
I have an idea.. could you make it 0.03 if coordinates are not activated and 0.06 if they are activated?

PR #75, opened 2018-04-06
Display winrate in separate box when in pondering mode.

bittsitt:
I added a winrate with bars that show black and whites winning chance, and more importantly, the change in this for the last move. This makes it easier to spot when you make a mistake (large negative change in win chance) and how "big" the mistake was. I considered adding color coding (orange for medium mistake, red for big mistake), but didn't add this now.

I didn't ask or make a feature request first, but I think this is useful.

image

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/analysis/Branch.java
+++ b/src/main/java/wagner/stephanie/lizzie/analysis/Branch.java
@@ -1,5 +1,6 @@
 package wagner.stephanie.lizzie.analysis;
 
+import wagner.stephanie.lizzie.Lizzie;
 import wagner.stephanie.lizzie.rules.Board;
 import wagner.stephanie.lizzie.rules.BoardData;
 import wagner.stephanie.lizzie.rules.Stone;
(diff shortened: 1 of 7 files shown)


dfannius:
While we're on the subject, one thing I have done in my local copy is to make the win rates on the board always be from Black's point of view, even when it is White's turn. This means that I don't have to keep subtracting from 100 in my head to see how it has changed from turn to turn, and matches the way that chess engines usually display their evaluation (+ is good for White, - is good for Black). I assume not everyone would like that change, but maybe we can make it configurable.

I wonder if this sort of display might look more elegant as a single vertical bar of fixed length, with the bottom 43% of it (in this case) colored black and the top 57% colored white. It should probably be an option in any case, as some people probably won't want to be distracted by it.

bittsitt:
One bar is a good idea. I was also thinking about a way to make a graphical representation of last move winrate change (which I think is the most important feature - was the last move a big mistake or not?) 

Maybe a red line that shows where it was before?

bittsitt:
Yeah, looks better with one bar. Is a line intuitive, or just confusing?

image

featurecat:
This looks really cool. I can't wait to merge it tomorrow :). Thank you. I'll mess around with the UI until I find something that looks best to me... and then ofc let other people customize it too lol. Maybe a couple options would be nice. I want to get creative with it!

cngoodboy:
Looks pretty good

bittsitt:
Warning, if anybody has pulled this, I did a force push.. Sorry about that

bittsitt:
I've added the winrate graph. Could maybe have been a separate PR, but since it was highly dependent on this branch, I just added it. Personally, I find it tremendously useful. It is jumping a bit up and down when you move back and forth, I suspect it is because the first winrates we get from leelaz after a move are either bad, or it is sending us winrates of previous move. Not sure which.

It would be natural to think about adding e.g. a prescan, where you load an sgf, then do prescan where you do a (limited) analysis of each move to generate the graph. Otherwise you have to step through everything to see what happened. Possible to expand on this, and I think people who use leela zero to analyze games will find this tremendously useful.

I'm currently also not displaying any variations. I think displaying all variations will just clutter too much, but showing the variation you are currently in could be interesting.

image

featurecat:
yay! maybe we can have a mistake highlighting

featurecat:
in the future

bittsitt:
With the graph, it is sort of obvious (large drop). :-)

PR #76, opened 2018-04-06
Y coordinates go from bottom to top per convention

### LINKED ISSUE #41: Y coordinate labeling is opposite of convention
While Go unfortunately does not have a totally standard coordinate system like chess does, it is conventional for the rows to be numbered 1-19 from the bottom to the top rather than from the top to the bottom. KGS, OGS, Crazy Stone, Leela all do it that way, for example.

dfannius:
Fixes #41.


isty2e:
Hmm, however two asian servers, namely Tygem and Fox, are using the opposite. This feature can be toggled perhaps?

dfannius:
Perhaps `c` can rotate between no coordinates, 1 at the bottom, and 1 at the top (and ideally the option can be saved). In the future if someone was sufficiently motivated we could even have [Japanese numbers](http://www5b.biglobe.ne.jp/~goban/g1go6f/igojuku/1.jpg) as an option.

PR #80, opened 2018-04-08
Implemented Copy SGF to clipboard and paste SGF from clipboard

aerisnju:
It is a very nice feature of Lizzie to load SGF from a file and save SGF to a file. But sometimes we would like to import/export a game from/to clipboard. For example, someone may paste the SGF contents of a game to a post and we are interested to the game, or we just want to copy/paste games from/to another game editor. I implemented this feature that allows copies from/pastes to clipboard.
Two new hot keys are added to support the feature, that is, Alt-C and Alt-V, to copy and paste. Because CTRL is already used in other functions. I resort to use Alt key.


featurecat:
Thanks. I don't see a good reason not to use CTRL though. it is used by other functions but there is no ctrl-c ctrl-c currently reserved, so I think I'll modify it to use those.

PR #91, opened 2018-04-14
Handicap/Komi and Save/Loading games - Part 2.

bvandenbon:
I went on testing and testing and testing and actually improved a lot of small things. e.g. in scenarios where you first start a handicap game, then change your mind and start a game without handicap after all, or increase/decrease handicap. And the use of handicap "1". ==> so a lot of small tweaks.

And then I made some adjustments for the error dialogs that show up when opening/closing/saving. In some of them the wrong constants were used. And e.g. when closing the window if you clicked cancel the window closed anyway. I made a lot of small tweaks for these scenarios.

I also made it show a warning dialog when you load a game with komi <> 7.5 and also when you start a game with komi <> 7.5. I think in both cases we should just let the user carry on. Because I can imagine people using this tool to write down their real-life games, and currently the new-game dialog is actually the only way to set the komi and handicap after all. Then secondly if you want to play handicap you do want to play with komi 0.5, right ? But leela really only supports 7.5.

Just 5 minutes ago I did a remerge of my fork with the master branch. So, I hope this helps to promote.

Kind regards, 
Have a nice weekend.
Bram Vandenbon


bvandenbon:
Now that I think about it.
The only supported komi by LeeLa is 7.5 I guess?
Or is it actually something else?

bvandenbon:
Then I probably did it right. Hope you like it.

PR #93, opened 2018-04-19
Don't respond to obsolete ponder output

### LINKED ISSUE #78: Stale leelaz data displayed
The following problem is made much more obvious by the way I personally use Lizzie: I have a local hack to always display winning percentages from Black's point of view, and my Leela Zero is a little laggy because I'm running it via `ssh` on a remote machine with a graphics card.

When `Board.place` is called, some information (like the new state of the board, and the current player) is updated immediately and rendered correspondingly, but we may still be waiting for the result of an old Leela Zero call to be coming in. `BoardRenderer` can then read an obsolete `bestMoves` list that no longer corresponds to the current board, and try to display it anyway. (In my situation, this causes old displays of, say, a winrate of 47% to be displayed as 53% for an instant before disappearing, since the current player has already been updated.)

I'm not certain what the best way to address this is. Maybe `Leelaz.playMove` could do something to make sure it doesn't try to parse old ponder output. The leelaz ponder output by itself doesn't seem to have enough information to identify what move produced it; otherwise we could identify and drop obsolete ponder output when we receive it.

dfannius:
Fixes #78.

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/analysis/Leelaz.java
+++ b/src/main/java/wagner/stephanie/lizzie/analysis/Leelaz.java
@@ -35,6 +35,9 @@ public class Leelaz {
     private boolean isPondering;
     private long startPonderTime;
 
+    // True while we're waiting for old ponder output to get flushed
+    private boolean isWaitingToStartPonder = false;
+
     // fixed_handicap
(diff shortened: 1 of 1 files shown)


featurecat:
Great! thanks

Issue #79, opened 2018-04-08
Multi-language support?


isty2e:
With the majority of Go players being asians, it seems natural to provide a language support for them.

Probably the in-application strings can be loaded from an external file, for the sake of ease of redistribution and user patches?

aerisnju:
It is easy to implement this by `ResourceBundle` in Java. I'll make an PR in a few days with translation to Chinese.

bittsitt:
What is best practice when you need to add a new string, add the english translation to all translation files, or only adding it to the english one?

aerisnju:
I think possibly adding it only to the english one is enough. It may be tedious to add it to all translations files manually, especially when there are more translations in the future. Possibly we need some scripts or tools to synchronize the english one with each of other translation files automatically. If someone would like to contribute to translation, he should use the scripts/tools first to synchronize the strings and then translate the strings added.

featurecat:
This is partially implemented. One problem we have right now is that translations get out of date when an English-speaking developer adds an English string but doesn't know the translation. We need to have a system that allows users to submit their own translations so that we can integrate them. Maybe we can read translations from a text file. That way users could change incorrect translations.

OlivierBlanvillain:
I think this could be closed after #97. To make sure that translations do not get out of date we should add a test that check that every key in the English l10n file is also present in the other files.

PR #97, opened 2018-04-19
Added localization support and Chinese translations

aerisnju:
This PR addresses #79. It changed some codes to use `ResourceBundle` in Java to display localized strings.

Before the change, a piece of code may display a hard-coded string:
```
displayString("Some hard coded string");
```

After the change, the code will use resource bundle to get localized string:
```
displayString(resourceBundle.getString("The.key.of.the.string"));
```
And in the base translation file:
```
The.key.of.the.string=Some hard coded string
```
In localized translation file:
```
The.key.of.the.string=......(Localized translations)
```

I made simplifiled Chinese translations too. But there's no translations of other languages. So Lizzie will display Chinese in a system with locale set to zh_CN, and display English by default in all other systems. If someone could help translation, he should copy the base file (`DisplayStrings.properties`) to a new file with the locale postfix(for example `DisplayStrings_jp.properties` for Japanese) and translate the strings.

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/gui/LizzieFrame.java
+++ b/src/main/java/wagner/stephanie/lizzie/gui/LizzieFrame.java
@@ -18,6 +18,7 @@ import wagner.stephanie.lizzie.rules.Board;
 import wagner.stephanie.lizzie.rules.SGFParser;
 
 import javax.imageio.ImageIO;
+import javax.swing.*;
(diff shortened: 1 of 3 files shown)


featurecat:
amazing. I'm sure our Chinese users will love it. I'll try to translate to Spanish when I have extra time.

featurecat:
(unless someone else is willing to, since I usually don't have time :C)

PR #103, opened 2018-04-21
Fix winrate graph

### LINKED ISSUE #102: Winrate graph drawing wrong data when in variation
I'm moving this out from #64 so I can write a little more without cluttering it up. This is part of number 7 on that checklist.

Currently, the winrate graph tries a little to be aware of the current variation (e.g., how long it is), but when it comes to drawing the lines it just uses the main trunk of the game tree. This causes incorrect data to be displayed, plus the data can extend over the board because its horizontal scaling is based on the variation's game length.

The simple way to fix this is

1. Fast-forward to the end of the current branch and get its last move number, for correct x-scaling (as is currently done)
2. Go backwards from that end node to the beginning, drawing the graph from right to left, instead of rewinding to the root and having to try to remember which branches to follow.

I'll do this myself today if no one gets to it first but I wanted to lay out the solution here in case someone else is working on it.

dfannius:
Fixes #102.

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/gui/WinrateGraph.java
+++ b/src/main/java/wagner/stephanie/lizzie/gui/WinrateGraph.java
@@ -48,25 +48,21 @@ public class WinrateGraph {
 
         g.setStroke(new BasicStroke(1));
 
-        // Go to bottom to find number of moves in main line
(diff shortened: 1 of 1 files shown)


featurecat:
perfect <3

PR #110, opened 2018-04-21
Navigate variation

bittsitt:
I've added support for navigating between variations using arrow keys. Basically, you move between moves with same move number.

There are two design choices I made that is good to be aware of.
- When moving to the right (variations), you are only allowed to move to variations that start on the same move on the main trunk. No such limitations when you're moving the other way. Possible to allow to move to any variation, but then you have to search the entire tree.
- If a variation is longer than the main branch (trunk), you are still allowed to move back to trunk

Disclaimer: There might be a more elegant way to do this (felt like a long complicated algorithm for something that "simple"), but I haven't spent any time looking for better solutions. Anyway, it works (I hope).


featurecat:
oops wrong number, this commit is not related.

PR #113, opened 2018-04-22
Change Assets to Resource Files in Jar Packages

cngoodboy:


featurecat:
why should we do this, though? having the assets outside the jar makes it easy for users to supply their own images to build their own themes.

cngoodboy:
Because I want to use a smaller number of files, but for what you said I can set a built-in CunstomTheme theme to load from the assets directory.

featurecat:
I think a lot of our users are not that technical but may want to edit the themes. I can't approve this PR because it would make editing the theme a lot harder compared to just replacing some files in ./assets.

PR #115, opened 2018-04-22
Always show main trunk winrate in graph, show variation as dashed line.

bittsitt:
For consideration. Better in principle, not sure if it works better in practice.


featurecat:
I think maybe we cannot add this to Lizzie 0.5 - what if the user makes a mistake and needs to undo? they will be stuck with the alternative graph for the rest of the game. We need to add some more variation tree controls before we can add this. Also, in a file with many plausible branches, I'm not sure if it's a good idea...

bittsitt:
I think it is more important to get 0.5 out, so if you are unsure, waiting is probably better.

If a user needs to undo, he will be stuck in a variation anyway. I see two options:
- A proper undo that deletes one or more moves
- A "convert current variation to main branch" command

The latter is a trivial algorithm, the former involves leelaz (I guess), but might be implemented already?

featurecat:
Please merge with next branch. WinrateGraph.java has conflicts.

bittsitt:
I've merged with next.

featurecat:
thanks!

PR #133, opened 2018-04-24
Only display visited nodes on winrate graph

### LINKED ISSUE #104: Winrate graph displays nonexistent and unreliable data
Currently the winrate graph draws a winrate for every node in the current variation, regardless of how many LZ visits that node has gotten. This data can sometimes be silly. Examples to trigger this behavior:

- Load a new SGF file (the whole game will exist but be unexplored)
- Hold down right-arrow to zoom through the loaded game (we get so few visits that the evaluation is nonsense)

I suggest that we require some minimum number of visits to a node before populating its winrate, and have a flag on the node (could be a boolean, could be a negative value of winrate) indicating whether its winrate is reliable and should be drawn.

dfannius:
Fixes #104. `BoardData` now includes the number of Leela Zero playouts for each node, and node stats are updated only when visited more than before. The winrate graph draws lines only where nodes have been visited. This means:

- You don't see long flat lines on the win graph where there is missing data (just loaded an SGF file, or quickly arrowed through part of it)
 - Quickly arrowing through nodes doesn't reset their winrate, which used to cause more long flat lines.

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/analysis/Branch.java
+++ b/src/main/java/wagner/stephanie/lizzie/analysis/Branch.java
@@ -22,7 +22,8 @@ public class Branch {
         Zobrist zobrist = board.getData().zobrist == null ? null : board.getData().zobrist.clone();
 
         // Dont care about winrate for branch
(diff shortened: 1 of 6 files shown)


featurecat:
great job

PR #157, opened 2018-05-02
Hotkey to hide variations etc. temporally

kaorahi:
When the board is too complicated with many best moves, I want to hide
variations etc. temporally.  This hotkey (z) works while it is held
down like "x" key.  So I selected "z", the next key in the keyboard.  I
am not sure which key is suitable, though.


featurecat:
Ok. sounds good. We have a lot of hotkeys now... hopefully we can start sorting them out in the near future!!

Issue #20, opened 2018-03-20
add score


igofun:
There's something I want to ask you
Add the score in the next version
key is "E"
please

isty2e:
Since LZ doesn't support scoring methods other than Tromp-Taylor, I wonder how practical this can be...?

CamWagner:
probably Chinese rules would give the same result so it's possible

PR #167, opened 2018-05-03
Scoring

bittsitt:
I added support for scoring at the end of the game (not estimate during game) #20 . As it is a fairly large change, I split it into multiple commits (even though most of them don't make much sense by themselves).

I also added display for captured stones. 

Click dead stones to mark as dead.


featurecat:
It seems like you may have pulled it from master instead of from next. There's some conflict with the next branch. Is it possible for you to merge them before I accept this PR?

bittsitt:
Might simply conflict with one of the previous PRs as well. I'll fix it.

bittsitt:
Will submit new PR, cleaner this way

Issue #164, opened 2018-05-03
Feature request: sweep to make graph


pondturtle:
I see that last version of lizzie shows graph of winrate development throughout the game, which is really cool feature. The issue is that the graph only build as I go through each of the moves stopping at each one so LZ has time to analyze them.

I think it would be great if there was option to let LZ go through the game first to build the graph in similar manner GRP does so I can quickly locate biggest winrate drops indicating mistakes. Thanks

dfannius:
It's a good idea and we have been discussing it! But it's good to have this open as a distinct issue to have a specific issue to refer to.

bittsitt:
I will re-submit the PR, but will wait for score-mode PR as I know I will have to fix some merge conflicts anyway.

bittsitt:
See #175. Hit a, enter number of playouts per move, sit back and wait.

featurecat:
#175 implements this. More work will be done on this in the future.

PR #168, opened 2018-05-03
Added simple pre-analysis mode. Hit a to start. Automatically moves f…

bittsitt:
Implements a simple version of #164. There might be other and better ways to do this, but this is a start.


featurecat:
Same request as #167, is it possible to merge with /next first? Looks like a great feature :)

bittsitt:
yes sure.. It just crashed on me here as well, looks like there is a bug that I need to look at also

PR #172, opened 2018-05-04
Fix slow convergence of winrate in small playouts

kaorahi:
This commit fixes slow convergence of winrate, though I do not
understand the reason :)

Without this commit, the dump_stats() between "~begin" and "~end"
tells too pessimistic values (e.g. 14%) when playouts are small (e.g. 1
or 2).  They cause slow convergence of winrate, that is serious for me
since my machine has no GPU.

I copied leelaz-src/UCTSearch.h in 8b0d598 from
https://github.com/featurecat/leela-zero
and added the line "bool ponder_helper();".

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,8 @@ any `leelaz_opencl_tuning` file you have into the directory.
 First, you will need to have a special version of Leela Zero that
 continually outputs pondering information. You can get this from one
 of the Lizzie releases or build it yourself; just substitute
-`leelaz-src/UCTSearch.cpp` for the `UCTSearch.cpp` file found in the
+`leelaz-src/UCTSearch.cpp` and `leelaz-src/UCTSearch.h` for the
+`UCTSearch.cpp` and `UCTSearch.h` files found in the
 Leela Zero sources. Then copy the resulting `leelaz` to the root of
 the Lizzie directory tree.
 
--- a/leelaz-src/UCTSearch.cpp
+++ b/leelaz-src/UCTSearch.cpp
@@ -618,6 +618,11 @@ int UCTSearch::think(int color, passflag_t passflag) {
 }
 
 void UCTSearch::ponder() {
+    while(ponder_helper()) {};                                      // lizzie
+}
+
+bool UCTSearch::ponder_helper() {                                   // lizzie
+    bool stopped_for_lizzie = false;                                // lizzie
     update_root();
 
     m_run = true;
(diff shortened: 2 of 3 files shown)


featurecat:
I don't understand what this changes/fixes. Could you describe it a bit more in detail? Sounds interesting.

kaorahi:
These are the screenshots with/without this commit.  The winrate graph
is much smoother with this commit on my slow machine.  I guess that
one may not notice this problem on a fast machine, because the graph
turns smoother as the playouts increase.

with_commit
without_commit

The procedure of both screenshots are as follows:
(1) Copy the following text to the clipboard.
    (;KM[7.5];B[pd];W[dp];B[pp];W[dd];B[sa];W[dj];B[ss];W[jd])
(2) Hit Ctrl-v key on lizzie (paste SGF).
(3) Wait for 3 seconds.
(4) Hit Down arrow key. (next move)
(5) Repeat (3) and (4).

    $ git log --oneline | head -n 1
    6ff5c53 Merge pull request #170 from bittsitt/scoremode
    $ ./leelaz -h | head -n 1
    Leela Zero 0.13  Copyright (C) 2017-2018  Gian-Carlo Pascutto and contributors
    $ uname -a
    Linux xxx 3.16.0-6-amd64 #1 SMP Debian 3.16.56-1 (2018-04-28) x86_64 GNU/Linux
    $ cat /proc/cpuinfo | grep 'model name'
    model name      : Intel(R) Core(TM) i7-4500U CPU @ 1.80GHz
    model name      : Intel(R) Core(TM) i7-4500U CPU @ 1.80GHz
    model name      : Intel(R) Core(TM) i7-4500U CPU @ 1.80GHz
    model name      : Intel(R) Core(TM) i7-4500U CPU @ 1.80GHz
    $ cat lizzie.properties | head -n 9
    {
      "leelaz": {
        "max-analyze-time-minutes": 60,
        "threads": 4,
        "max-game-thinking-time-seconds": 16,
        "weights": "network",
        "gpu": [],
        "print-comms": false
      },

bittsitt:
@kaorahi Great idea, but hopefully proper GTP support in Leela zero will happen quite soon (I came up with some ideas on how to extend GTP, and it seems to have been accepted - maybe we will see something within the next week?).  In that case, it might be better to submit this patch to leelea zero if the problem is still there. 

Although if there are no side effects for those with GPUs (I haven't tested it), there is no harm to add it I guess.

featurecat:
so you are pausing the pondering every 0.1 seconds instead of outputting the data mid-ponder? I don't understand why it would be different. is that correct? is our current code for Leelaz wrong?

featurecat:
I'll do some testing to compare, when I get the time. This week I think.

featurecat:
with https://github.com/gcp/leela-zero/pull/1388 it seems this is a virtual loss issue. we should try to fix this issue that way, as suggested in the thread, I think!

featurecat:
thoughts

bittsitt:
Agree. Makes sense to fix this in LZ..

jkiliani:
I opened a PR for this: https://github.com/gcp/leela-zero/pull/1412. It's possible though that gcp will prefer another approach to fix it, we'll see when he reviews PRs next...

featurecat:
It's been merged! We will use the Leela Zero next branch for our next release then :)

PR #175, opened 2018-05-05
Add simple auto-analysis mode. Hit a to start.

bittsitt:
Trigger auto-analysis by hitting 'a'. Starts from current position, stop with a, stops when it reaches end of game.


kaorahi:
"break;" for "case VK_Z:" is missing in Input.java.  Anyway, this
auto-analysis mode is very convenient and I enjoy it well :-)

Issue #119, opened 2018-04-23
stable/dev branch


Ka-zam:
Perhaps it would be helpful to add a branch for development and then merge back into stable when deemed fully tested? Tags could be added for releases.

Ka-zam:
I just mean that you wouldn't necessarily need to tag every merge to master with a tag.

bittsitt:
@featurecat : You added the next branch (which I've started using for PR), but my last three PRs were merged to master instead of next. Was that just a mistake on your side? Just wondering which branch to base my PRs on... :-)

featurecat:
Oops no idea how that happened. I'll fix. thanks.

featurecat:
will fix in morning... (11 hrs from now)!

bittsitt:
I have a pretty good idea as I've done it a couple of times myself ;-)

bittsitt:
@featurecat  Ugh.. I just discovered what happened, and it is all my fault. My next branch was created off master branch (next and master were at the same head so I didn't notice). Sorry about that!!! I've resubmitted score-mode, should be on next now. 

But, the pass_bug_fix ended up on master. Should also be on next (that is just one line, so easy to merge/cherry pick). Otherwise, people will experience strange things when trying out score mode

PR #180, opened 2018-05-06
Fixes exception when stepping past a pass move

bittsitt:
As mentioned in #119, this has already been merged on master, but we need it in next as well.


featurecat:
Thanks for catching this problem in next branch

Issue #178, opened 2018-05-06
Use Leela 0.11.0 on Lizzie


kaorahi:
What do you think about using Leela 0.11.0 instead of Leela Zero with
Lizzie UI?  Leela 0.11.0 is still useful for analysis of handicapped
games because Leela Zero overlooks bad moves and says "B will win
anyway" until the last mistake of B.

I have a toy implementation here.
  https://github.com/kaorahi/lizzie/tree/leela_nonzero
Is there any possibility of merging if I brush it up?

It works in a similar way as #172.  Namely, it repeats
(1) Send "time_left b 0 0" to start pondering.
(2) Wait for 1 second.
(3) Send "name" to stop pondering.
(4) Read outputs of leela.
(5) Go to (1).

See README_leela0110.txt in the above branch for trying it.  As a side
effect, it can also run the original Leela Zero without modification.

featurecat:
yes. this is a good idea. we can also use this as a failsafe for incompatible Leelaz versions.

PR #182, opened 2018-05-06
Failsafe mode for incompatible Leela

kaorahi:
I implemented #178 as a failsafe mode for both Leela 0.11.0 and the
original Leela Zero without modification.  See the bottom of README.md
to use Leela 0.11.0.

Though "parseLineNormal" and "parseLineFailSafe" have many common
lines, I dare to copy them naively because "parseLineNormal" will be
rewritten for the official pondering output by the original Leela zero
in near future.
(cf.) https://github.com/gcp/leela-zero/pull/1242

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,3 +46,16 @@ To build the code and package it:
 
 (or whatever the current version of the shaded `jar` file is in
 `target/`).
+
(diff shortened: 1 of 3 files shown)


featurecat:
Exciting :)

PR #184, opened 2018-05-07
option to display advatage in number of handicap stones instead of winrate.

typohh:
As networks vary, the relative winrate differ too given the same position. Weak network might only give 60% winrate to white if black passes as first move, some recent leela networks give 80% winrate for the same, and elf will give 99% winrate. For this reason, quoting a network winrate is meaningless unless you also specify version of a network. However its possible to normalize winrates by converting them to number of handicap stones, which are independent of the network strength and meaningful comparison across different networks. So instead of saying a position has 80% winrate, one could say instead that the position has 0.4 handicap stone advantage and conversely 20% winrate would be -0.4 handicap stones.  This PR calculates the normalization factor for a network, and then applies the conversion when  displaying advantage in handicap stones on the board. The old way of displaying is still possible with Config.


src/main/java/wagner/stephanie/lizzie/Lizzie.java
```diff
@@ -27,20 +29,53 @@ public static void main(String[] args) throws IOException, JSONException, ClassN
...
         }).start();
 
-        PluginManager.loadPlugins();
-
-        board = new Board();
+        
+    }
 
-        frame = new LizzieFrame();
+    public static void estimatePassWinrate() {
```
featurecat: these methods probably belong somewhere in the analysis package. Leelaz.java would be ok

src/main/java/wagner/stephanie/lizzie/Lizzie.java
```diff
@@ -27,20 +29,53 @@ public static void main(String[] args) throws IOException, JSONException, ClassN
 
         config = new Config();
 
+        PluginManager.loadPlugins();
```
featurecat: Why was the order of these changed?

src/main/java/wagner/stephanie/lizzie/Lizzie.java
```diff
@@ -27,20 +29,53 @@ public static void main(String[] args) throws IOException, JSONException, ClassN
...
+    	board.clear();
+    }
+    
+    public static double mHandicapWinrate=25;
+    
+    /**
+     * Convert winrate to handicap stones, by normalizing winrate by first move pass winrate (one stone handicap).
+     */
+    public static double winrateToHandicap(double pWinrate) {
+        return Math.signum(0.5-(pWinrate/100))*Math.log(1-Math.abs(1-(pWinrate/100)*2))/-Math.log(1-Math.abs(1-(mHandicapWinrate/100)*2));
```
featurecat: these should be on a few lines, with named variables so that we can see how it works. Not all on one line. A brief comment explaining how it works would be helpful. (slightly more detailed than the docstring comment)

src/main/java/wagner/stephanie/lizzie/Lizzie.java
```diff
@@ -27,20 +29,53 @@ public static void main(String[] args) throws IOException, JSONException, ClassN
...
+    	while( stats.totalPlayouts < 1 ) {
+    		try {
+				Thread.sleep(100);
+			} catch (InterruptedException e) {
+				throw new Error(e);
+			}
+    		stats=leelaz.getWinrateStats();
+    	}
+    	mHandicapWinrate=stats.maxWinrate;
+    	System.out.println("handicap winrate set to " + stats.maxWinrate + " after " + stats.totalPlayouts + " playouts." );
```
featurecat: should remove this debug line

src/main/java/wagner/stephanie/lizzie/Lizzie.java
```diff
@@ -27,20 +29,53 @@ public static void main(String[] args) throws IOException, JSONException, ClassN
...
 
-        PluginManager.loadPlugins();
-
-        board = new Board();
+        
+    }
 
-        frame = new LizzieFrame();
+    public static void estimatePassWinrate() {
+    	leelaz.playMove(Stone.BLACK, "A1"); // we use A1 instead of pass, because networks have some experience with this due to early randomness but none with pass. hence the network has reasonably converged and don't need lots of playouts for reasonable estimate.
```
featurecat: could you please reformat this long comment, and also add a comment explaining how we might test the correctness of Black A1 vs pass vs black having two handicap stones etc.

src/main/java/wagner/stephanie/lizzie/Lizzie.java
```diff
@@ -27,20 +29,53 @@ public static void main(String[] args) throws IOException, JSONException, ClassN
...
+    	leelaz.togglePonder();
+    	WinrateStats stats=leelaz.getWinrateStats();
+    	while( stats.totalPlayouts < 1 ) {
+    		try {
+				Thread.sleep(100);
```
featurecat: fix indentation. Should be equivalent to 4 characters wide. This is 8, followed by 16+8 on the next line.

bittsitt:
How accurate is this? Have you tested it across multiple engines/networks at different phases of a game? 

I like the idea of normalizing, but I fear handicap stones are less intuitive and can be confusing, especially to weaker/new players. Everybody understands %. Nobody will understand -1.2. Would it be better to convert it back to % instead? 

What really would be great was a proper score estimate.  I've been thinking of trying to add a layer to the LZ network to do score estimation, but I don't have a computer with decent GPU to train it on.

Interestingly enough, I did a quick test now on a very close game that B won by 0.5. With no more points to play on the board, LZ thinks W has a winrate of 57% with 20k playouts...

src/main/java/wagner/stephanie/lizzie/Lizzie.java
```diff
@@ -27,20 +29,53 @@ public static void main(String[] args) throws IOException, JSONException, ClassN
...
-
-        board = new Board();
+        
+    }
 
-        frame = new LizzieFrame();
+    public static void estimatePassWinrate() {
+    	leelaz.playMove(Stone.BLACK, "A1"); // we use A1 instead of pass, because networks have some experience with this due to early randomness but none with pass. hence the network has reasonably converged and don't need lots of playouts for reasonable estimate.
+    	leelaz.togglePonder();
+    	WinrateStats stats=leelaz.getWinrateStats();
```
featurecat: I haven't released a coding style guide (yet), but please put spaces on either side of = for your assignments: eg
WinrateStats stats = leelaz.getWinrateStats();

A code reformatter such as Eclipse Ctrl+Shift+F or Intellij Ctrl+Alt+L will do this for you automatically. I prefer Intellij's but either will help.

src/main/java/wagner/stephanie/lizzie/Config.java
```diff
@@ -149,6 +149,7 @@ private JSONObject createDefaultConfig() {
         ui.put("show-next-moves", true);
         ui.put("win-rate-always-black", false);
         ui.put("confirm-exit", false);
+        ui.put("handicap-instead-of-winrate",true);
```
featurecat: I have reservations about making this the default option, since no go program has ever done this before to my knowledge. It's less familiar to most users. Please change this to 
ui.put("handicap-instead-of-winrate", false);

I'm happy to add this is a menu option, when we have an options menu inside Lizzie. I think it's a really cool feature.

src/main/java/wagner/stephanie/lizzie/Lizzie.java
featurecat: A docstring comment would be helpful, to describe the output of this function.

featurecat:
@bittsitt please keep me updated on your score estimator. LZ being able to say "white is 5.5 points ahead" would be a lot more intuitive than saying "White has a 62% winrate".

src/main/java/wagner/stephanie/lizzie/Lizzie.java
typohh: there was no dependencies that break because of this order, but with the original order I had no good place to run the estimatePassWinrate() without potential race condition or additional synchronization due to requirement of board and leelaz

featurecat: So long as there's a reason, that's fine. In the future we should remove the interdependencies of leelaz / lizzieFrame since they really should be separate components.

kaorahi:
Interesting. The winrate graph looks different with this nonlinear conversion.
https://github.com/kaorahi/lizzie/tree/wrate2hcap
orig
conv

PR #204, opened 2018-05-11
Support for gib file format used by Tygem

bittsitt:
Supports players names, komi, moves, pass and handicap stones.


featurecat:
super useful, thanks :)

PR #217, opened 2018-05-14
Print all lines from leelaz if "print-comms" is true

kaorahi:
I hope to watch all lines from leelaz for debugging (e.g. #216).
Do you have a reason to hide some lines?


featurecat:
Does this also include the ~begin and ~end lines? (or in future: the info lines)? The reason I omitted them was because they're quite spammy.

kaorahi:
I see. I found that I can use leelaz's option --logfile for my purpose.

PR #222, opened 2018-05-15
Play the best move by "," key if no variation is shown

kaorahi:
I feel that it is consistent to play "leelaz's suggestion" by "," key in both cases.

* Play the displayed variation (#159), or...
* Play the best move if no variation is displayed (this patch).

Both will be useful for examining leelaz's suggestion deeper.

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/gui/BoardRenderer.java
+++ b/src/main/java/wagner/stephanie/lizzie/gui/BoardRenderer.java
@@ -103,6 +103,19 @@ public class BoardRenderer {
     }
 
     /**
+     * Return the best move of Leelaz's suggestions
+     *
+     * @return the coordinate name of the best move
+     */
+    public String bestMoveCoordinateName() {
+        if (bestMoves == null || bestMoves.size() == 0) {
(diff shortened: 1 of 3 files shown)


bittsitt:
I like the feature, but personally, I think I would prefer that "," just plays one move (current best move). When it plays the whole variation, you jump directly to the end result, but you already see the end result by hovering the mouse. By just moving one and one instead, you can quickly play it out which makes it easier to follow. Just my 2 cents?

BTW, you've added many nice features, I especially love the hover in winrate graph feature!! It is awesome!

kaorahi:
Hi, @bittsitt san. Congratulations and thanks for your effort on PR to leelaz.

I intended to watch leelaz's suggestion *after* the displayed variation by "," key in #159.
For watching the process of the displayed variation, I implemented another feature #223.

BTW, I've canceled the hover in winrate graph in #218. I think drag is better.

@featurecat san, sorry for my too many pull requests.
Most of my wishes have been granted now.

featurecat:
it's been great I also appreciate all the PRs you've made :) more is better.

Issue #84, opened 2018-04-11
Multiple rings


TFiFiE:
There is currently a red ring around the most visited child. Maybe also have a blue one for the best-scoring one and make it purple when they're the same. I also think they could be a bit thicker.

featurecat:
what if we gave them a special color besides green?

TFiFiE:
That's also possible, but you might want to start thinking about making all those colors configurable, including those on the green-to-red spectrum, because they are especially notorious from a colorblind perspective.

dfannius:
Note that in master there's also a black or white ring showing the next move if there is one, so we have to be extra careful about clutter. (My personal opinion is that the current best scoring move is easy enough to find by eye and doesn't need additional calling out.)

TFiFiE:
New idea: If the most visited child is also the best-scoring one, give it a fully blue color. Otherwise make it cyan and give the best-scoring one a blue ring. This is only two more colors and is also better from a theoretical point of view, because the best-scoring child won't then lose the visual cue indicating how far it's behind in terms of visits.

featurecat:
IIRC, Leela Zero often abandons nodes with seemingly higher winrates than the principal variation, leaving them mostly unexplored. Possibly it's because it found a good refutation and stopped searching? But the winrate doesn't really get updated. Would this be an issue?

dfannius:
Yeah, I don't really understand this behavior of LZ and I would like to. It's not really obvious to me what it means when LZ has a node with a better winrate than some others but it is not exploring it as much, and I'm a little reluctant to call out "this looks like the best option but LZ is declining to explore it" moves.

In minimax or alpha-beta, if you find a refutation, the value gets propagated up the tree appropriately and you don't get a situation where the eval that you're publishing doesn't match what you really think of the move. So I'm a little confused here.

featurecat:
Apparently the actual winrate displayed by LeelaZ is an averaged value of some sort of its children nodes. I don't know the logic behind that vs the alpha-beta style. Roy7 has a fork that probably behaves more like alpha-beta: https://github.com/gcp/leela-zero/pull/883. It might be stronger, too, but it's been a pull request for a long time. It probably would fix the winrate problem.

featurecat:
roy7 do you have any idea when+if your branch will get merged?

roy7:
[...]

PR #230, opened 2018-05-19
Return of the special rings

TFiFiE:
As the variant of #84 I've settled on, this will give the most visited child move a red ring if it does not also have the highest average win rate and hands out a blue ring in the inverse case.
Although I personally found this useful to investigate whether these two metrics always eventually agree on the best move, I can totally understand it if this doesn't get merged because it's too much in terms of visual information.


featurecat:
I will have to test this out. I can't really find the time to test until after Wednesday. could you attach a screenshot?

featurecat:
(or anyone can attach it if they happen to have one)

kaorahi:
Screenshots are pasted in #241.
The ring gives precious information on convergence of search.

featurecat:
ok. thanks. Your pr looks really innovative :)

PR #241, opened 2018-05-24
Add subboard that always shows the principal variation

kaorahi:
I feel that zero-click UI is the key invention of Lizzie.  It makes
analyses of games much fun.  I had never imagined that the difference
between 1-click and 0-click is so large until I experienced it by
myself.

Along this direction, I added a subboard that always shows the
principal variation.  We do not need even mouse-hovering to see it.
It is hidden by shift-Z key (no-hint view in #223) and enlarged by
mouse click on it.  I also added "show-subboard" and "large-subboard"
into lizzie.properties.

Of course this feature is entirely redundant.  But I think that "can
do" and "can do with no action" is different.

Though this PR is rather large, half of changed lines are mere layout
parameters.

subboard1
subboard2

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/Config.java
+++ b/src/main/java/wagner/stephanie/lizzie/Config.java
@@ -19,6 +19,7 @@ public class Config {
     public boolean showBestMoves = true;
     public boolean showNextMoves = true;
     public boolean showSubBoard = true;
+    public boolean largeSubBoard = false;
     
     public JSONObject config;
     public JSONObject persisted;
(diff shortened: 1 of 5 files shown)


bittsitt:
Nice idea, but it would look much nicer if the boards had the same color.

featurecat:
I'll merge this now and worry about the graphics later... I can see there might be some problems getting fancy-mode to work, so for now this is fine. We can also adapt the basic graphics to use whatever color the fancy-mode graphics are when averaged out.

featurecat:
It looks very good :)

PR #242, opened 2018-05-25
Revert "Use non-fancy style for subboard"

kaorahi:
This reverts commit f89e83619e5aabf139107af84f9f776d13084093.

I intentionally used non-fancy style for subboard because it is "sub".
But I know my design sense is unreliable.  This patch reverts the
commit f89e836 (Use non-fancy style for subboard) and applies the same
style for both boards.  Is it better?  Either is ok for me.


featurecat:
Yeah I think this is better. We can see if users have a performance impact but it should be fine.

PR #247, opened 2018-05-26
Support for new leela zero GTP format. 

bittsitt:
Took a while because of a bit of back and forth on LZ (plus I've been a bit busy), but I think we're finally there. Now works with standard LZ (not yet released, so only next branch of LZ).

### DIFF
--- a/src/main/java/wagner/stephanie/lizzie/Config.java
+++ b/src/main/java/wagner/stephanie/lizzie/Config.java
@@ -148,6 +148,7 @@ public class Config {
         leelaz.put("max-analyze-time-minutes", 2);
         leelaz.put("max-game-thinking-time-seconds", 2);
         leelaz.put("print-comms", false);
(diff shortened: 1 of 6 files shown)


kaorahi:
Congratulations!

We also need to delete the last section of README.md if we drop failsafe mode.
Though we lose Leela 0.11 support, I understand that any "mode" increases complexity
and makes maintenance harder.

featurecat:
omg thanks!! I'd been hoping to do this sooner but I'm moving tomorrow. with this I think we could actually release Sunday night (eastern time) or Monday :). 

all that's left to do that could be contributed by others is automatic network downloading. if anyone feels like doing this, feel free to include extra files to include in the release (such as unzip utilities). but I am planning to do this anyway :) 

you're the best :)

featurecat:
Leela 0.11 support might be nice to add back in the future. I'm not sure. it depends on what other features are added to lizzie that may depend on Leela zero / custom Leela zero code

PR #257, opened 2018-05-30
No need to gunzip network file. leelaz (>=0.14) accepts *.gz

kaorahi:


featurecat:
interesting! useful :D thanks

PR #272, opened 2018-06-02
Fix issues on komi and handicap in SGF reader

kaorahi:
These commits fix the following issues in SGF reader.

1. GameInfo was lost for handicapped SGF.
2. Reading failed for SGF with empty "KM[]".

example
http://gokifu.net/t2.php?s=7001484524613847


featurecat:
thanks! make sure to pull from next, not from master.

PR #281, opened 2018-06-04
Pick a reasonable default leelaz binary

apetresc:
Now that Leela-Zero has the new GTP extensions that Lizzie requires, it's no longer mandatory to compile a custom leelaz binary, which means it's no longer a hard requirement for there to be a leelaz binary in Lizzie's current working directory.

To make life easier for packagers (like me), when constructing the default config.txt, Lizzie should examine the PATH and find the system-wide leelaz. Of course, a local ./leelaz should still take priority (even if `.` is not actually on your PATH).

This makes that the default behaviour. Works for me on a fresh install with leela-zero installed in /usr/bin.


bittsitt:
Nice one, but you should request the merge to be into next instead of master (just click edit on top)...

I just added a pull request (#284) before I saw this one. My PR also needs to check full path, so either @featurecat should fix it when applying my PR, or I can do it once this one has been merged. Probably easiest to merge this first

apetresc:
@bittsitt Good call, I missed that lizzie uses `next` the same way leela-zero does. Rebased all my PRs, thanks :)

PR #282, opened 2018-06-04
Fix parsing of /network-profiles page

apetresc:
Sometime recently, http://zero.sjeng.org/network-profiles changed their page, the links are no longer hard-coded in the HTML but instead populated dynamically in Javascript. So the old regex was broken and just never found a match.

Fix the regex to now grab the first hash from the inline JSON that Vue uses to populate the table. Works on my own manual testing.


featurecat:
I would rather use zero.sjeng.org/best-network-hash here, suggested by roy7. The network profiles page may change syntaxes again.

apetresc:
Excellent idea, I didn't know about `/best-network-hash`. Implemented and tested :)

featurecat:
wow that is a lot simpler haha. Great! thanks :)

PR #283, opened 2018-06-04
Package assets as part of the JAR

apetresc:
In my continuing effort to make Lizzie relocatable, the assets should be included as part of the JAR, rather than requiring the user to run Lizzie from inside a directory that has the /assets folder. This accomplishes that.

Tested and works locally. With the other PRs filed separately, you can now run the Lizzie JAR in `/usr/share/java` from any other directory (or a `.desktop` file).


featurecat:
OK. I was thinking earlier that this may be too much added effort to re-theme the program for some users. I've rejected this kind of change before. but on second thought, it is cleaner and probably better to keep the assets in the package. we will need to notify users that the assets folder outside the package is now useless, if they have one.

PR #287, opened 2018-06-05
Enhancement: Window size and maximization

Ka-zam:
This PR allows setting persistent window size and whether to start in maximized mode.


featurecat:
thanks. I don't think you meant to check in lizzie.sh, just be careful.

PR #289, opened 2018-06-05
Fix too slow "End", "Ctrl-V", etc. by omitting unnecessary lz-analyze

kaorahi:
"Ctrl-V" takes several minutes on my machine.
It is caused by "lz-analyze" after every "play ..." and "undo".
You may not notice this issue if your machine is fast.

The first patch introduces a command queue and defers sending commands
until leelaz responds to the previous command.
Unnecessary "lz-analyze", that will be stopped immediately by the next command,
is removed from the queue before sending it to leelaz.

I also add the second patch for preventing possible hang-up in unexpected cases
that cannot happen theoretically.

    $ git log --oneline | head -n 1
    1912f4a Merge pull request #272 from kaorahi/komi_sgf

### DIFF
--- a/src/main/java/featurecat/lizzie/analysis/Leelaz.java
+++ b/src/main/java/featurecat/lizzie/analysis/Leelaz.java
@@ -33,6 +33,7 @@ public class Leelaz {
     private long maxAnalyzeTimeMillis;//, maxThinkingTimeMillis;
     private int cmdNumber;
     private int currentCmdNum;
+    private ArrayDeque<String> cmdQueue;
 
     private Process process;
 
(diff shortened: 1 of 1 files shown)


kaorahi:
Because...

(1) It is not restricted to VK_END and VK_V.  There are many commands
that need such treatments, and more commands will be added in future.
I could not list all such commands, but this queue fixes them at once.

(2) Lizzie has to remember the pondering status before the command and
restore it afterward if we use togglePonder-strategy.  We then need
more code in addition to simple togglePonder().

(3) This queue supports auto-repeated keys naturally.  Namely,
pondering is stopped when we begin to hold down up-arrow key, and
pondering is restarted immediately when we release up-arrow key.  If
we use togglePonder-strategy, we need additional code for the
KeyReleased event to realize such a smart behavior.

featurecat:
I will review this a little more before merging.

PR #297, opened 2018-06-13
Fix "Did not find network/weights"

kaorahi:
I get the below alert dialog at startup without this patch.

    Did not find network/weights.
    Update config.txt (network-file) or download from Leela Zero homepage

The value of the variable startfolder should be ".". But I get "./leelaz".


featurecat:
oh that was probably my mistake - I should have tested better.

PR #326, opened 2018-07-14
Allow file loading using CLI argument

OlivierBlanvillain:
The Leela 11 GUI from https://sjeng.org/leela.html has the same behavior, UNIX users might prefer that way of opening files over the Java dialog.

(I hope I'm not spamming too much with the 3 PRs, they are pretty unrelated so I figured I should open separate ones)


featurecat:
no it's good I prefer the 3. thank you. you also fixed the filename extension issue. does it still work with .gib files?

OlivierBlanvillain:
> you also fixed the filename extension issue. does it still work with .gib files?

I'm not sure what issue you are referring too, both .gib and .sgf work fine for me on master and next, and they can both be loaded from the CLI after this PR.

The commit where I touched `saveSgf` and `openSgf` is only renaming these methods to a more accurate name that reflect the fact that they handle both .sgf and .gib!

PR #357, opened 2018-09-03
 Fix location checks, and use engine-start-location directly

### LINKED ISSUE #323: File location checking is broken on next branch, works on master
`next` branch in `Config.java` supports looking at PATH to auto-detect `leelaz` program. However in `Lizzie.java` and `Leelaz.java` there is logic that checks `startfolder` which uses completely different logic.

infinity0:
Fixes #323 

Don't pointlessly recalculate paths, just take them from the config directly since they've already been calculated by then.

Drop `engine-program` since it's not being used, just take the first element of the split string `engine-command` like we already do for `ProcessBuilder`.

### DIFF
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -105,15 +105,7 @@ public class Config {
 
         // Check engine configs
         JSONObject leelaz = config.getJSONObject("leelaz");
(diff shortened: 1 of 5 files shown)


featurecat:
sorry, you're right. I will approve this pr soon, or also we will probably have some new contributors soon who can also approve prs. within 1 week

PR #363, opened 2018-09-17
Fix #362: show best moves after --visit exhaustion

OlivierBlanvillain:
This PR adds a new parser for what I've called "summary line". Leelaz will systematically output small summaries like the following once it reached the maximum number of visits specified with the `--visit` flag:

```
  O4 ->       5 (V: 53.21%) (N: 39.95%) PV: O4 F3 C6 C17
  F3 ->       4 (V: 54.48%) (N: 34.51%) PV: F3 F17 E17 F16
```

Parsing this data fixes #362.

This PR also setups JUnit and adds a few tests for the new regular expression.

Issue #356, opened 2018-09-02
About Multiple Engine Support & Heights Switching


zsalch:
Hi

I am looking forward to supporting more engines in the next release of the lizzie, such as AQ, Phoenix, Elf, etc. I am prefer to AQ because it uses Japanese rules, and most of AI only support Chinese rules. AQ can make up for this deficiency.

But I also know that supporting a new engine requires more work, so I temporarily made a simple solution that allows Lizzie to switch between multiple weights, which also helps me see the analysis status between different weights.

Because I have not do full tests, so only submitting the changes in my branch. Those who are interested can refer to them(https://github.com/zsalch/lizzie/tree/next_dev).

Change logs:

1. In the config.txt file, add a configuration (engine-command-list) to add a new command lines:

    "engine-command": "./leelaz --gtp --lagbuffer 0 --weights %network-file --threads 2",
    "engine-command-list": [
      "./leelaz --gtp --lagbuffer 0 --weights leelaz_172_b1c --threads 2",
      "./leelaz --gtp --lagbuffer 0 --weights human_best --threads 2",
      "./leelaz --gtp --lagbuffer 0 --weights elf_1_62b --threads 2",
      "./leelaz --gtp --lagbuffer 0 --weights leelaz_157_d35 --threads 2"
    ],

2. Use the shortcut key Ctrl+number to switch the weight and display the weight name in the left down:
    Ctrl+0: Default switch to "engine-command"
    Ctrl+(1-9): Switch to "engine-command-list"

3. If the current node is in Branch, it also supports keeping the current node when switching weights.

The effect is as following:

elf 2 d13:
![image](https://user-images.githubusercontent.com/42595514/44957152-37280d80-af01-11e8-9fe7-704f776d77f4.png)
leela zero 172 b1c:
![Uploading image.png…]()

jar file:

[Uploading Lizzie Heights Switch.zip…]()

wonsiks:
Wow~ this is great~~~
Good job~

wonsiks:
I saw that there're more configs in the config file.
Are they working at the moment?
What about show-dynamic-komi option? How can I use it?

apetresc:
[...]

PR #364, opened 2018-09-18
next: Support Read/Write Sgf with Variation, AW/AB & Comment

### LINKED ISSUE #350: About display comment of sgf and read/write sgf with variation & Winrate.
First of all, I am very grateful for the amazing program.

It has helped me a lot.

However, I hope that can see the comments of the sgf when review the game, and I also hope to save the variation and Winrate.

I guess the author must be very busy, I can't wait and to try to make some minor changes.

1. Add the display the comment of the sgf (C tag): Forgive me very lazy, directly displayed under the variation panel, please ignore the ugly style :(
2. Add the ability to read and write the sgf with variation
3. By the way, write the Winrate in the comment of sgf for each step (C tag): Just a little try

Attached the codes and files, the code base on the release 0.5, and use the class (WrapString) of the Jim Menard. Thanks Jim Menard.
There are not many changes, not list the differences.

Just a reference for those in need.

Sorry for use the Issue because I can't find the author E-mail.
If not suitable, please remove it.

Looking forward to related functions of next release.

Thanks again.

zsalch:
Closes #350 

Support following features when Read/Write the SGF file:

1. Variations tree
2. AW/AB (Add White)/(Add Black) stone
3. Comment (C tag)

### DIFF
--- a/src/main/java/featurecat/lizzie/rules/SGFParser.java
+++ b/src/main/java/featurecat/lizzie/rules/SGFParser.java
@@ -324,9 +324,8 @@ public class SGFParser {
         // *  format: ";B[xy]" or ";W[xy]"
         // *  with 'xy' = coordinates ; or 'tt' for pass.
 
-
         // Write variation tree
-        builder.append(generateNode(board, history.nextNode()));
+        builder.append(generateNode(board, history.getCurrentHistoryNode()));
 
         // close file
         builder.append(')');
@@ -357,19 +356,19 @@ public class SGFParser {
                 if (data.comment != null) {
                     builder.append(String.format("C[%s]", data.comment));
                 }
+            }
 
-                if (node.numberOfChildren() > 1) {
-                    // Variation
-                    for (BoardHistoryNode sub : node.getNexts()) {
-                        builder.append("(");
-                        builder.append(generateNode(board, sub));
-                        builder.append(")");
-                    }
-                } else if (node.numberOfChildren() == 1) {
-                    builder.append(generateNode(board, node.next()));
-                } else {
-                    return builder.toString();
+            if (node.numberOfChildren() > 1) {
+                // Variation
+                for (BoardHistoryNode sub : node.getNexts()) {
+                    builder.append("(");
+                    builder.append(generateNode(board, sub));
+                    builder.append(")");
                 }
+            } else if (node.numberOfChildren() == 1) {
+                builder.append(generateNode(board, node.next()));
+            } else {
+                return builder.toString();
             }
         }
 
--- a/src/main/java/featurecat/lizzie/rules/SGFParser.java
+++ b/src/main/java/featurecat/lizzie/rules/SGFParser.java
@@ -72,7 +72,7 @@ public class SGFParser {
         Map<Integer, Integer> subTreeStepMap = new HashMap<Integer, Integer>();
         // Comment of the AW/AB (Add White/Add Black) stone
         String awabComment = null;
-		// Previous Tag
+        // Previous Tag
         String prevTag = null;
         boolean inTag = false, isMultiGo = false, escaping = false;
         String tag = null;
@@ -86,7 +86,6 @@ public class SGFParser {
 
         String blackPlayer = "", whitePlayer = "";
 
-        PARSE_LOOP:
         // Suppoert unicode charactors (UTF-8)
         for (int i = 0; i < value.length(); i++) {
             char c = value.charAt(i);
@@ -104,10 +103,10 @@ public class SGFParser {
                         // Initialize the step count
                         subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(0));
                     } else {
-                    	if (i > 0) {
-                    	    // Allow the comment tag includes '('
-                    	    tagContentBuilder.append(c);
-                    	}
+                        if (i > 0) {
+                            // Allow the comment tag includes '('
+                            tagContentBuilder.append(c);
+                        }
                     }
                     break;
                 case ')':
@@ -152,8 +151,8 @@ public class SGFParser {
                         if (move == null) {
                             Lizzie.board.pass(Stone.BLACK);
                         } else {
-                        	// Save the step count
-                        	subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue() + 1));
+                            // Save the step count
+                            subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue() + 1));
                             Lizzie.board.place(move[0], move[1], Stone.BLACK);
                         }
                     } else if (tag.equals("W")) {
@@ -161,17 +160,17 @@ public class SGFParser {
                         if (move == null) {
                             Lizzie.board.pass(Stone.WHITE);
                         } else {
-                        	// Save the step count
-                        	subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue() + 1));
+                            // Save the step count
+                            subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue() + 1));
                             Lizzie.board.place(move[0], move[1], Stone.WHITE);
                         }
                     }  else if (tag.equals("C")) {
-                    	// Support comment
-                    	if ("AW".equals(prevTag) || "AB".equals(prevTag)) {
-                    		awabComment = tagContent;
-                    	} else {
-                    		Lizzie.board.comment(tagContent);
-                    	}
+                        // Support comment
+                        if ("AW".equals(prevTag) || "AB".equals(prevTag)) {
+                            awabComment = tagContent;
+                        } else {
+                            Lizzie.board.comment(tagContent);
+                        }
                     } else if (tag.equals("AB")) {
                         int[] move = convertSgfPosToCoord(tagContent);
                         if (move == null) {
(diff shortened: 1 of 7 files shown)


src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -96,14 +101,27 @@ private static boolean parse(String value) {
...
+                    	    // Allow the comment tag includes '('
+                    	    tagContentBuilder.append(c);
+                    	}
                     }
                     break;
                 case ')':
                     if (!inTag) {
-                        subTreeDepth -= 1;
                         if (isMultiGo) {
-                            break PARSE_LOOP;
```
OlivierBlanvillain: You can also remove the `PARSE_LOOP` label (outside the parsing loop)

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -79,12 +87,9 @@ private static boolean parse(String value) {
         String blackPlayer = "", whitePlayer = "";
 
         PARSE_LOOP:
-        for (byte b : value.getBytes()) {
-            // Check unicode charactors (UTF-8)
-            char c = (char) b;
-            if (((int) b & 0x80) != 0) {
-                continue;
-            }
+        // Suppoert unicode charactors (UTF-8)
```
OlivierBlanvillain: Typo Suppoert

OlivierBlanvillain: But I think you should remove that comment. After your changes the code supports unicode because it uses charAt instead of looking that bytes, so it will be hard to understand the comment out of context.

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -96,14 +101,27 @@ private static boolean parse(String value) {
                 case '(':
                     if (!inTag) {
                         subTreeDepth += 1;
+                        // Initialize the step count
+                        subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(0));
+                    } else {
+                    	if (i > 0) {
```
OlivierBlanvillain: Can you explain the `i > 0` here? It seems that if we are inside a tag we always want to append that `'('`

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -96,14 +101,27 @@ private static boolean parse(String value) {
...
+                    	}
                     }
                     break;
                 case ')':
                     if (!inTag) {
-                        subTreeDepth -= 1;
                         if (isMultiGo) {
-                            break PARSE_LOOP;
+                            // Restore to the variation node
+                            for (int s = 0; s < subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue(); s++) {
```
OlivierBlanvillain: It's a bit suspicious to recompute `subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue()` at every step of the loop. Any reason for doing that instead of computing that just outside of the for?

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -134,15 +152,26 @@ private static boolean parse(String value) {
                         if (move == null) {
                             Lizzie.board.pass(Stone.BLACK);
                         } else {
+                        	// Save the step count
+                        	subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue() + 1));
```
OlivierBlanvillain: Indentation is off here and a bit all over the PR... Could you configure your text editor to output spaces instead of tabs? The output should look nicer on GitHub and it would also be more consistent with the rest of the lizzie code base.

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -134,15 +152,26 @@ private static boolean parse(String value) {
...
                             Lizzie.board.place(move[0], move[1], Stone.BLACK);
                         }
                     } else if (tag.equals("W")) {
                         int[] move = convertSgfPosToCoord(tagContent);
                         if (move == null) {
                             Lizzie.board.pass(Stone.WHITE);
                         } else {
+                        	// Save the step count
+                        	subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(subTreeStepMap.get(Integer.valueOf(subTreeDepth)).intValue() + 1));
```
OlivierBlanvillain: This line seams to be copy pasted. Is a way to avoid the duplication?

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -96,14 +101,27 @@ private static boolean parse(String value) {
                 case '(':
                     if (!inTag) {
                         subTreeDepth += 1;
+                        // Initialize the step count
+                        subTreeStepMap.put(Integer.valueOf(subTreeDepth), Integer.valueOf(0));
```
OlivierBlanvillain: I think in Java it's not required to explicitly convert between int and Integer, the java compiler will automatically box and unbox `int` and `Integers` to make things work. Concretely this line could be simplified to `subTreeStepMap.put(subTreeDepth, 0);`

This comment applies for the entirety of this PR, it would be nice if you could go over the diff and remove all the `Integer.valueOf()` and `.intValue()` calls that would be automatically inserted, I think the result would be simpler to read!

OlivierBlanvillain:
@zsalch These are some really good additions! The only thing that worries be is the absence of tests. The code seams to be rather low level and fragile, and it very hard just by looking at the diff to make sure that you didn't break anything. Also, how is this code going to evolve in the future? In the absence of tests, it would very easy for someone to accidentally break the support for comments of variations without noticing.

My suggestion to move forward would be to add integration tests to this PR. I would like this parsing code to be accompanied with a set of sample games that demonstrate all the features handled by the parser, and a few tests to show that the parser doesn't crash on these games and produces the expected output.

You could maybe start from this commit https://github.com/featurecat/lizzie/pull/363/commits/efc8fe31e9d6cba3a12f351b66ab847be6c34932 where I setup JUnit for the project, and write a simple test for the parser. If you need help with that I might also be able to contribute a test sample to kick start the effort.

Also I think I will wait for this PR to be ready to merge before looking at #365.

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -250,65 +285,92 @@ private static void saveToStream(Board board, Writer writer) throws IOException
...
+        // Write variation tree
+        builder.append(generateNode(board, writer, history.nextNode()));
 
         // close file
         builder.append(')');
         writer.append(builder.toString());
     }
+
+    // Generate node
+    private static String generateNode(Board board, Writer writer, BoardHistoryNode node) throws IOException {
```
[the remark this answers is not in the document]
zsalch: Garbage code, deleted.

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -250,65 +285,92 @@ private static void saveToStream(Board board, Writer writer) throws IOException
                     builder.append(String.format("[%c%c]", x, y));
                 }
             }
+        } else {
+        	// Process the AW/AB stone
```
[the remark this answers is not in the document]
zsalch: When has handicap, there is Black stones only, the original code already is processed.
So just add a else.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Removed.

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -250,65 +285,92 @@ private static void saveToStream(Board board, Writer writer) throws IOException
...
+
+        // Write variation tree
+        builder.append(generateNode(board, writer, history.nextNode()));
 
         // close file
         builder.append(')');
         writer.append(builder.toString());
     }
+
+    // Generate node
```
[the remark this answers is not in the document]
zsalch: Fixed.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: For skip first '(' that at the top of the Sgf file.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Sorry, I have ignored this problem because with multiple editors.
Fixed it.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: You are right.
The previous IDE compiler reported error, now there is no such problem.
Fixed.

zsalch:
@OlivierBlanvillain 

Thank you for your review, so the code is much clearer.

Introducing JUnit is a good idea and I will add some JUnit use cases.

For the test variations and comments, I think sgf e-book will be a good choice, of course, we have just added the AB / AW and C tag processing, for other tags have yet to be added.

I will upload some sgf for testing, but most of them are in Chinese. Maybe you can provide some English sgf files.

Regarding the test points, I think the first thing is the following:

* 1. Whether AB/AW and C tag and variations are correctly stored in the NodeList;
* 2. Is the AB/AW placed in the correct position;
* 3. Whether the variations are displayed correctly;
* 4. The comment for each step is displayed correctly;
* 5. When saving, AB/AW and C tag and branches are saved correctly.

Since the structure of the Node and the NodeList is not modified, the general operation should be normal when variations are correctly displayed. But this also requires testing to verify.

Before do JUnit testing, you may be want to refer #356 . Here are some simple test cases.

zsalch:
@OlivierBlanvillain 
I have added two simple test cases. Please check.

PR #365, opened 2018-09-18
next: Support display the comment of the SGF file

zsalch:
Closes:    #350 

This pull request based on #364 

1. Support for displaying comments of the sgf files.
    Default is close, can press key 'T' to toggle, and add a item 'show-comment' in the config file.
    image

2. Allow use a hidden item 'comment-font-size' in the config to forced specify the comment font size, for example:
    "comment-font-size": 20

### DIFF
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -16,6 +16,7 @@ public class Config {
     public boolean showWinrate = true;
     public boolean showVariationGraph = true;
     public boolean showComment = false;
+    public int commentFontSize = 0;
     public boolean showRawBoard = false;
     public boolean showCaptured = true;
     public boolean handicapInsteadOfWinrate = false;
@@ -134,6 +135,7 @@ public class Config {
         showWinrate = uiConfig.getBoolean("show-winrate");
         showVariationGraph = uiConfig.getBoolean("show-variation-graph");
         showComment = uiConfig.optBoolean("show-comment", false);
+        commentFontSize = uiConfig.optInt("comment-font-size", 0);
         showCaptured = uiConfig.getBoolean("show-captured");
         showBestMoves = uiConfig.getBoolean("show-best-moves");
         showNextMoves = uiConfig.getBoolean("show-next-moves");
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -15,6 +15,7 @@ public class Config {
     public boolean showMoveNumber = false;
     public boolean showWinrate = true;
     public boolean showVariationGraph = true;
+    public boolean showComment = false;
     public boolean showRawBoard = false;
     public boolean showCaptured = true;
     public boolean handicapInsteadOfWinrate = false;
@@ -132,6 +133,7 @@ public class Config {
         showBranch = uiConfig.getBoolean("show-leelaz-variation");
         showWinrate = uiConfig.getBoolean("show-winrate");
         showVariationGraph = uiConfig.getBoolean("show-variation-graph");
+        showComment = uiConfig.optBoolean("show-comment", false);
         showCaptured = uiConfig.getBoolean("show-captured");
         showBestMoves = uiConfig.getBoolean("show-best-moves");
         showNextMoves = uiConfig.getBoolean("show-next-moves");
@@ -178,6 +180,9 @@ public class Config {
     }
     public void toggleShowVariationGraph() {
         this.showVariationGraph = !this.showVariationGraph;
+    }
+    public void toggleShowComment() {
+        this.showComment = !this.showComment;
     }
     public void toggleShowBestMoves() {
         this.showBestMoves = !this.showBestMoves;
--- a/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
+++ b/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
@@ -21,7 +21,6 @@ import featurecat.lizzie.rules.GIBParser;
 import featurecat.lizzie.rules.SGFParser;
 import org.json.JSONObject;
 import org.json.JSONArray;
-import org.json.JSONException;
 
 import javax.swing.*;
 import javax.swing.filechooser.FileNameExtensionFilter;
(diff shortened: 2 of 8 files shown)


featurecat:
@OlivierBlanvillain could you review this one? Thanks! Looks so useful :)

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -945,4 +1014,37 @@ public void pasteSgf() {
...
+     *
+     * @param g
+     * @param x
+     * @param y
+     * @param w
+     * @param h
+     * @param full
+     * @return
+     */
+    private int drawCommnet(Graphics2D g, int x, int y, int w, int h, boolean full) {
```
OlivierBlanvillain: Typo: `s/drawCommnet/drawComment/`

src/main/java/featurecat/lizzie/rules/BoardHistoryList.java
```diff
@@ -88,6 +88,20 @@ public BoardData next() {
         return head.getData();
     }
 
+    /**
+     * moves the pointer to the right, returns the node stored there
+     *
+     * @return the next node, null if there is no next node
+     */
+    public BoardHistoryNode nextNode() {
```
OlivierBlanvillain: Unused?

OlivierBlanvillain:
Could you rebase this PR on top of next? When I reached these lines https://github.com/featurecat/lizzie/pull/364/files#diff-d7874015cfce6652bd9e035733aaaef1R89 I realized I already reviewed parts of this code...

zsalch:
@OlivierBlanvillain 
I have reset the commits for a clean commit tree and fixed the code.
Please help to check. Thanks.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -945,4 +1014,37 @@ public void pasteSgf() {
...
+     * @param full
+     * @return
+     */
+    private int drawComment(Graphics2D g, int x, int y, int w, int h, boolean full) {
+        String comment = (Lizzie.board.getHistory().getData() != null && Lizzie.board.getHistory().getData().comment != null) ? Lizzie.board.getHistory().getData().comment : "";
+        int cHeight = full ? h : (int)(h*0.5);
+        int fontSize = (int)(Math.min(getWidth(), getHeight()) * 0.98 * 0.03);
+        try {
+            fontSize = Lizzie.config.uiConfig.getInt("comment-font-size");
+        } catch (JSONException e) {
```
OlivierBlanvillain: This seems out of place to me. I feel that in this part of the code we should not have to think about JSON or default value, instead simply access something like `Lizzie.config.uiConfig.commentFontSize`, the rest should be handled as it is for other config options.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -945,4 +1014,37 @@ public void pasteSgf() {
...
+            if (fontSize < 16) {
+                fontSize = 16;
+            }
+        }
+        Font font = new Font(systemDefaultFontName, Font.PLAIN, fontSize);
+        commentPane.setFont(font);
+        commentPane.setText(comment);
+        commentPane.setSize(w, cHeight);
+        createCommentImage((comment != null && !comment.equals(this.cachedComment)) ? true : false, w, cHeight);
```
OlivierBlanvillain: `expr ? true : false` should be the same as `expr`...

src/main/java/featurecat/lizzie/gui/VariationTree.java
```diff
@@ -79,7 +79,7 @@ public void drawTree(Graphics2D g, int posx, int posy, int startLane, int maxpos
         g.setColor(curcolor);
 
         // Draw main line
-        while (cur.next() != null && posy + YSPACING < maxposy) {
+        while (cur.next() != null && ((posy + YSPACING + DOT_DIAM) < maxposy)) { // Fix oval cover issue sometimes
```
OlivierBlanvillain: Can you explain a bit more what this is fixing?

zsalch: Sometimes the variation node covers the comment panel.
But this bug is previous version. I'll test current version for this bug.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
zsalch: Wow, what is the code! Removed.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
zsalch: Changed to config.

src/main/java/featurecat/lizzie/gui/VariationTree.java
zsalch: It seems no bug now. Restored.

src/main/java/featurecat/lizzie/Config.java
```diff
@@ -134,6 +135,7 @@ public Config() throws IOException {
         showWinrate = uiConfig.getBoolean("show-winrate");
         showVariationGraph = uiConfig.getBoolean("show-variation-graph");
         showComment = uiConfig.optBoolean("show-comment", false);
+        commentFontSize = uiConfig.optInt("comment-font-size", 0);
```
OlivierBlanvillain: Default to `16` instead of `0`?

zsalch: The 16 is not a default value, just a minimum value.
Now remove it, if the font size is not set, the font size automatically changes with the window size.

OlivierBlanvillain: If you want 16 as minimum that logic should probably be restored here https://github.com/featurecat/lizzie/pull/365/files#diff-c51e6645e657a3c7c444daafb9ad21e0R1031

zsalch: I think it may be does not need.

OlivierBlanvillain: I think we need either a sensible default of a minimum value. If we have neither of these and a user happens to have a config file without the `comment-font-size` setting then he won't be able to see the comment text.

zsalch: Ok, I have restored the minimum size 16.

OlivierBlanvillain:
Looks good, thanks!

PR #367, opened 2018-09-18
next: Support Display Only Last Move Number

zsalch:
The change supports display only last move number that can be specified in the config.txt.
For example, only display last 10 move number:

"ui": {
    "only-last-move-number": 10,

image

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/BoardRenderer.java
+++ b/src/main/java/featurecat/lizzie/gui/BoardRenderer.java
@@ -470,7 +470,7 @@ public class BoardRenderer {
         int[] moveNumberList = branch == null ? Lizzie.board.getMoveNumberList() : branch.data.moveNumberList;
 
         // Allow to display only last move number
-        int lastMoveNumber = Lizzie.board.getData().moveNumber;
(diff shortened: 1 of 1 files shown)


src/main/java/featurecat/lizzie/gui/BoardRenderer.java
```diff
@@ -469,11 +469,20 @@ private void drawMoveNumbers(Graphics2D g) {
 
         int[] moveNumberList = branch == null ? Lizzie.board.getMoveNumberList() : branch.data.moveNumberList;
 
+        // Allow to display only last move number
+        int lastMoveNumber = Lizzie.board.getData().moveNumber;
+        int onlyLastMoveNumber = (!Lizzie.config.uiConfig.isNull("only-last-move-number")) ? Lizzie.config.uiConfig.getInt("only-last-move-number") : 9999;
```
OlivierBlanvillain: What about `uiConfig.optInt("only-last-move-number", 9999)` instead?

zsalch: Better method. I have changed and committed.

PR #378, opened 2018-10-03
next:  The custom theme allow to use external image, set font & color by config

zsalch:
Base on #365 

Add a new way to customize the theme by external images, setup font & color of the comment.

The configuration method is as following:
    1) Change the 'theme' value to specify your theme name:
        "theme": "<your-theme-name>",
        for example:
        "theme": "mylike",
    3) Create subfolders in the lizzie folder
        <lizzie-folder>/theme/<your-theme-name>
        for example:
        <lizzie-folder>/theme/mylike
    4) The custom-theme folder must include the config file 'theme.txt', the format as followings(use default when option not be specified):
{
  "background-image" : "background.png",
  "board-image" : "board.png",
  "black-stone-image" : "black.png",
  "white-stone-image" : "white.png",
  "font-name": "Wawati SC",
  "solid-stone-indicator": false,
  "comment-background-color": [0, 0, 0, 200],
  "comment-font-color": [255, 255, 255, 255],
  "winrate-line-color": [0, 255, 0, 255],
  "winrate-miss-line-color": [0, 0, 178, 255],
  "blunder-bar-color": [255, 0, 0, 150]
}
    e) Will use default theme when not found related config item.
    f) Run the Lizzie.


ParmuzinAlexander:
It is hard make random stone "white-stone-image" : "1.png,10.png" ?

zsalch:
@ParmuzinAlexander 
The random stone is not particularly difficult to implement, but I want to know more about the scenario, which is to randomly select a stone at each move, or choose one at the beginning of the game.
If each move is randomly chosen, it needs to cache more types of stones, performance may have a little impact and require more logic.

ParmuzinAlexander:
@zsalch 
Each move. Or maybe not random but in sequence 1.png,2.png,....10.png,1.png

OlivierBlanvillain:
@zsalch Can you rebase this one now that #365 is merged?

zsalch:
@OlivierBlanvillain 
No problem. But now ITheme&DefaultTheme has been removed, does this mean that CustomTheme needs to be adjusted accordingly, after all, it is based on ITheme & DefaultTheme.

OlivierBlanvillain:
Yes. My suggestion would be to always have custom theme on. Basically add a few options to the config JSON to change colors and the paths to the various images we use.

zsalch:
Ok, I'll remove some redundant paths and use 'theme' instead of 'theme'+'custom-theme' in the config.txt file.

zsalch:
@OlivierBlanvillain 
I have updated the code and commit again.

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,187 @@
...
+    this.themeName = themeName;
+    this.path = this.pathPrefix + File.separator + this.themeName;
+    File file = new File(this.path + File.separator + this.configFile);
+    if (file.canRead()) {
+      FileInputStream fp;
+      try {
+        fp = new FileInputStream(file);
+        config = new JSONObject(new JSONTokener(fp));
+        fp.close();
+      } catch (FileNotFoundException e) {
```
OlivierBlanvillain: `FileNotFoundException` extends `IOException`, so you can remove that line.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -137,6 +141,13 @@ public LizzieFrame() {
     JSONArray windowSize = Lizzie.config.uiConfig.getJSONArray("window-size");
     setSize(windowSize.getInt(0), windowSize.getInt(1)); // use config file window size
 
+    // Allow change font in the config
+    if (theme.fontName() != null) {
+      systemDefaultFontName = theme.fontName();
```
OlivierBlanvillain: I think these name are obsolete if you start changing them. Instead I would use

- `fontName`
- `regularFont`
- `boldFont`

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -137,6 +141,13 @@ public LizzieFrame() {
...
+      OpenSansRegularBase = new Font(systemDefaultFontName, Font.PLAIN, 12);
+      OpenSansSemiboldBase = new Font(systemDefaultFontName, Font.BOLD, 12);
```
OlivierBlanvillain: Should the front size also be configurable? I it often the case than it needs adjustment depending on the font...

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -440,9 +451,11 @@ public void paint(Graphics g0) {
       if (Lizzie.leelaz != null && Lizzie.leelaz.isLoaded()) {
         if (Lizzie.config.showStatus) {
           String pondKey = "LizzieFrame.display." + (Lizzie.leelaz.isPondering() ? "on" : "off");
-          String pondText = resourceBundle.getString(pondKey);
```
OlivierBlanvillain: It seems that this part of the diff comes from #393, it will probably disappear if you do another merge...

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,187 @@
...
+
+/** Theme Allow to load the external image & theme config */
+public class Theme {
+  BufferedImage blackStoneCached = null;
+  BufferedImage whiteStoneCached = null;
+  BufferedImage boardCached = null;
+  BufferedImage backgroundCached = null;
+
+  private String themeName = null;
+  private String configFile = "theme.txt";
```
OlivierBlanvillain: To be honest I don't really like splitting the configuration into multiple files, I think it makes things more complicated / confusing. What do we gain by allowing users to have multiple `theme.txt` as opposed to copy past their entire `config.txt` into `config-theme-1.txt` ... `config-theme-N.txt`?

Also, I don't think the average user would often switch between several theme, I believe the average user is going to tweak once upon install the program and never change the graphics again. Putting everything in the main `config.txt` file should make things simpler for this the use case.

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,187 @@
+package featurecat.lizzie.theme;
+
+import java.awt.Color;
```
OlivierBlanvillain: Since you use it a bunch of times in this file:

```java
import static java.io.File.separator;
```

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  public BufferedImage blackStone() {
+    if (blackStoneCached == null) {
+      try {
+        blackStoneCached =
+            ImageIO.read(
+                new File(
+                    this.path
```
OlivierBlanvillain:
```
String p = this.path + /* File. */separator + config.optString("black-stone-image", "black.png");
```

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,187 @@
...
+    return blackStoneCached;
+  }
+
+  public BufferedImage whiteStone() {
+    if (whiteStoneCached == null) {
+      try {
+        whiteStoneCached =
+            ImageIO.read(
+                new File(
+                    this.path
```
OlivierBlanvillain: Same, just to make it fit in the 100 char limit...

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,187 @@
...
+          e1.printStackTrace();
+        }
+      }
+    }
+    return backgroundCached;
+  }
+
+  /** Use custom font */
+  public String fontName() {
+    return config.optString("font-name", null);
```
OlivierBlanvillain: Default to `null`? Why not return the actual default font name?

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * The background color of the comment panel
+   *
+   * @return
+   */
+  public Color commentBackgroundColor() {
+    JSONArray color = config.optJSONArray("comment-background-color");
+    if (color != null) {
+      return new Color(color.getInt(0), color.getInt(1), color.getInt(2), color.getInt(3));
```
OlivierBlanvillain: Maybe we should give a nice error message in case somebody messes up the config file and `color.length != 4`. Same comment for the methods below.

zsalch:
@OlivierBlanvillain 
I still prefer to use the independent theme settings, so if you need to share a theme, just copy the corresponding directory file, and simply specify the theme name in the config.txt file.

However, I also agree that you can set theme-related options directly in the config.txt file, so that users can modify them without enabling the theme.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
zsalch: Different areas have different font sizes. If need to configure them, you may need to add several different configurations.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
zsalch: Fixed.

zsalch:
@featurecat @OlivierBlanvillain 
We've added some new configuration options, and we need to determine which ones are enabled by default.
Such as: show-blunder-bar, show-comment etc.

OlivierBlanvillain:
@zsalch Alright, if you think the split in `config.txt`/`theme.txt` makes more sense, I'm OK with it.

I think we should turn both of these on by default for better discoverability.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
zsalch: @featurecat @OlivierBlanvillain 

Currently the font usage is now as follows:

OpenSansSemiboldBase: winrate text on the stone
OpenSansRegularBase: some UI text
systemDefaultFontName: some prompt text

And now the font size is basically responsive with the window, if you want to use the configuration file to change, I think we need to take some time to design the font usage.

So now only three font configurations are defined. The font size option only supports comment.

OpenSansSemiboldBase: winrate-font-name
OpenSansRegularBase: ui-font-name
systemDefaultFontName: font-name

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,147 @@
...
+    BufferedImage boardCached = null;
+    BufferedImage backgroundCached = null;
+
+    private String themeName = null;
+    private String configFile = "theme.txt";
+    private String pathPrefix = "theme";
+    private String path = null;
+    private JSONObject config = new JSONObject();
+
+    public Theme(String themeName) {
```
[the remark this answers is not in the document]
zsalch: Removed.

src/main/java/featurecat/lizzie/theme/Theme.java
zsalch: Removed.

src/main/java/featurecat/lizzie/theme/Theme.java
```diff
@@ -0,0 +1,147 @@
...
+        this.themeName = themeName;
+        this.path = this.pathPrefix + File.separator + this.themeName;
+        File file = new File(this.path + File.separator + this.configFile);
+        if (file.canRead()) {
+            FileInputStream fp;
+            try {
+                fp = new FileInputStream(file);
+                config = new JSONObject(new JSONTokener(fp));
```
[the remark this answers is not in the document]
zsalch: As conversation, keep two ways.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
zsalch: Rename to fontName, uiFontName, winrateFontName.

src/main/java/featurecat/lizzie/theme/Theme.java
zsalch: As conversation, keep two ways.

src/main/java/featurecat/lizzie/theme/Theme.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/theme/Theme.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/theme/Theme.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/theme/Theme.java
zsalch: Change fontName logic.

Issue #308, opened 2018-06-25
There is one suggestion.


kjh2671:
How about add a function to record the winning rate and visit count to sgf?
I think it will help people review again later.

featurecat:
yes... this is a top requested feature. I am not working on it at the moment, but it is a top priority. lately I don't have much free time to work on lizzie. it depends on the contributions of others or when I have time to do it myself.

PR #368, opened 2018-09-18
next: Support Switch Multiple Engine Heights

zsalch:
Closes: #356 

Support to switching multiple engine heights. (Currently only support Leela Zero Engine)

How to config:
1. In the config.txt file, add a itme (engine-command-list) to add some new command lines:

    "engine-command": "./leelaz --gtp --lagbuffer 0 --weights %network-file --threads 2",
    "engine-command-list": [
      "./leelaz --gtp --lagbuffer 0 --weights leelaz_172_b1c --threads 2",
      "./leelaz --gtp --lagbuffer 0 --weights human_best --threads 2",
      "./leelaz --gtp --lagbuffer 0 --weights elf_1_62b --threads 2",
      "./leelaz --gtp --lagbuffer 0 --weights leelaz_157_d35 --threads 2"
    ],

2. Use the shortcut key Ctrl+number to switch the weight and display the weight name in the left down, the engine command line will be displayed in the Title:

    Ctrl+0: Default switch to "engine-command"
    Ctrl+(1-9): Switch to "engine-command-list"

3. If the current node is in Branch, it also supports keeping the current node when switching weights.

Demo:
image
image

PR #379, opened 2018-10-03
next: Allow write winrate etc. to the comment of the sgf by append-winrate-to-comment

zsalch:
Closes: #308 
Base on: #368 

Support save the winrate etc. into the comment of Sgf (need turn on "append-winrate-to-comment" in the config.txt)
format as followings:

Move \<move number>
\<Winrate> (\<Last Move Rate Difference>)
(\<Weight name> / \<Playouts>)

for example:
W[dd]C[Move 2
53.5% (-0.1%)
(elf_v1 / 18k playouts)

other comments];

### DIFF
--- a/src/main/java/featurecat/lizzie/rules/BoardData.java
+++ b/src/main/java/featurecat/lizzie/rules/BoardData.java
@@ -72,6 +72,9 @@ public class BoardData {
    */
   public void addProperty(String key, String value) {
     SGFParser.addProperty(properties, key, value);
+    if ("N".equals(key) && comment.isEmpty()) {
+      comment = value;
+    }
   }
 
   /**
--- a/src/main/java/featurecat/lizzie/rules/SGFParser.java
+++ b/src/main/java/featurecat/lizzie/rules/SGFParser.java
@@ -4,10 +4,12 @@ import static java.util.Arrays.asList;
 
 import featurecat.lizzie.Lizzie;
 import featurecat.lizzie.analysis.GameInfo;
+import featurecat.lizzie.analysis.Leelaz;
 import java.io.*;
 import java.text.SimpleDateFormat;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -307,6 +309,11 @@ public class SGFParser {
             "KM[%s]PW[%s]PB[%s]DT[%s]AP[Lizzie: %s]",
             komi, playerW, playerB, date, Lizzie.lizzieVersion));
 
+    // To append the winrate to the comment of sgf we might need to update the Winrate
+    if (Lizzie.config.appendWinrateToComment) {
+      Lizzie.board.updateWinrate();
+    }
+
     // move to the first move
     history.toStart();
 
@@ -398,6 +405,11 @@ public class SGFParser {
         // Node properties
         builder.append(data.propertiesString());
 
+        if (Lizzie.config.appendWinrateToComment) {
+          // Append the winrate to the comment of sgf
+          data.comment = formatComment(node);
+        }
+
         // Write the comment
         if (!data.comment.isEmpty()) {
           builder.append(String.format("C[%s]", data.comment));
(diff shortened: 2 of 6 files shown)


src/main/java/featurecat/lizzie/rules/BoardData.java
```diff
@@ -72,6 +72,9 @@ public static BoardData empty(int size) {
    */
   public void addProperty(String key, String value) {
     SGFParser.addProperty(properties, key, value);
+    if ("N".equals(key) && comment.isEmpty()) {
```
OlivierBlanvillain: Should the line above be moved in an else branch?

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -307,6 +308,17 @@ private static void saveToStream(Board board, Writer writer) throws IOException
             "KM[%s]PW[%s]PB[%s]DT[%s]AP[Lizzie: %s]",
             komi, playerW, playerB, date, Lizzie.lizzieVersion));
 
+    // For append the winrate to the comment of sgf, maybe need to update the Winrate
```
OlivierBlanvillain: *To* append the winrate to the comment of sgf *we might* need to update the Winrate

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -307,6 +308,17 @@ private static void saveToStream(Board board, Writer writer) throws IOException
             "KM[%s]PW[%s]PB[%s]DT[%s]AP[Lizzie: %s]",
             komi, playerW, playerB, date, Lizzie.lizzieVersion));
 
+    // For append the winrate to the comment of sgf, maybe need to update the Winrate
+    if (Lizzie.config.appendWinrateToComment) {
+
+      // Update winrate
+      Leelaz.WinrateStats stats = Lizzie.leelaz.getWinrateStats();
```
OlivierBlanvillain: After this change this code would be duplicated in 5 places, place / nextMove / nextMove (overloaded) / previousMove. Could you please factor this out into an `updateWinrate` method to get ride of all this duplication?

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -421,6 +438,62 @@ private static String generateNode(Board board, BoardHistoryNode node) throws IO
...
+      if (Lizzie.config.handicapInsteadOfWinrate) {
+        double currHandicapedWR = Leelaz.winrateToHandicap(100 - curWR);
+        double lastHandicapedWR = Leelaz.winrateToHandicap(lastWR);
+        lastMoveDiff = String.format(": %.2f", currHandicapedWR - lastHandicapedWR);
+      } else {
+        lastMoveDiff = String.format("(%.1f%%)", 100 - lastWR - curWR);
+      }
+    }
+
+    // Format:
```
OlivierBlanvillain: Duplicate information with the javadoc above

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -421,6 +438,62 @@ private static String generateNode(Board board, BoardHistoryNode node) throws IO
...
+   */
+  private static String formatComment(BoardHistoryNode node) {
+    BoardData data = node.getData();
+    String engine = Lizzie.leelaz.currentWeight();
+
+    // Playouts
+    String playouts = Lizzie.frame.getPlayoutsString(data.playouts);
+
+    // Last winrate
+    BoardData lastNode = node.previous().get().getData();
```
OlivierBlanvillain: How do you know that it's safe to call `.get()` here?

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -421,6 +438,62 @@ private static String generateNode(Board board, BoardHistoryNode node) throws IO
...
+    boolean validLastWinrate = (lastNode != null && lastNode.playouts > 0);
```
OlivierBlanvillain: `lastNode` cannot be `null`.

src/main/java/featurecat/lizzie/rules/SGFParser.java
```diff
@@ -421,6 +438,62 @@ private static String generateNode(Board board, BoardHistoryNode node) throws IO
...
+      }
+    }
+
+    // Format:
+    // Move <Move number> <Winrate> (<Last Move Rate Difference>)
+    // (<Weight name> / <Playouts>)
+    String wf = "Move %d\n%s %s\n(%s / %s playouts)";
+    String nc = String.format(wf, data.moveNumber, curWinrate, lastMoveDiff, engine, playouts);
+
+    if (data.comment != null) {
```
OlivierBlanvillain: `comment` cannot be null, you probably want `!data.comment.isEmpty()` here.

src/main/java/featurecat/lizzie/rules/BoardData.java
zsalch: No, I want to keep the N attribute.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Fixed

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Yes, fixed.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Removed

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/rules/SGFParser.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/rules/SGFParser.java
OlivierBlanvillain: I see. But still, it's dangerous to has assumptions across methods like this without documentation. If you really don't won't to handle the case where a node doesn't have a previous node, the I would add it to the javadoc and start the method witha a `if (!node.previous().isPresent()) throw new IllegalArgumentException("...")`.

zsalch: Ok, fixed.

featurecat:
I'm merging, because we need to hurry things along for the 0.6 release.

Issue #273, opened 2018-06-02
ponder data no longer being read during play mode - won't update winrate graph as playing


featurecat:
when playing against LZ ('n' or 'enter'), the winrate graph window is no longer updated

featurecat:
mostly as a reminder to myself to fix this.

OlivierBlanvillain:
Is this fixed by #363?

featurecat:
maybe. I'll check this later

PR #382, opened 2018-10-04
next: fixed wrong winrate when playing a handicap game

zsalch:
Fixed wrong winrate when playing a handicap game: #273 #363 
image


src/main/java/featurecat/lizzie/gui/Input.java
```diff
@@ -196,7 +196,8 @@ public void keyPressed(KeyEvent e) {
       case VK_SPACE:
         if (Lizzie.frame.isPlayingAgainstLeelaz) {
           Lizzie.frame.isPlayingAgainstLeelaz = false;
-          Lizzie.leelaz.togglePonder(); // we must toggle twice for it to restart pondering
+          // It seems does not need
```
OlivierBlanvillain: Then just remove it...

zsalch: Ok, removed.

zsalch:
When the handicap game is playing, the bestMoves is not initialized, and the previous information is retained, so we need to initialize and refresh the latest information.

src/main/java/featurecat/lizzie/analysis/Leelaz.java
```diff
@@ -329,6 +332,8 @@ private void parseLine(String line) {
         } else if (isThinking && !isPondering) {
           if (Lizzie.frame.isPlayingAgainstLeelaz) {
             Lizzie.board.place(params[1]);
+            togglePonder();
+            isPondering = false;
```
OlivierBlanvillain: Why manually set `isPondering` here? `togglePonder()` seems to already be doing the correct thing depending on the value of that boolean. It appears that setting it by hand puts us in an inconsistent state where we are pondering but the `isPondering` says otherwise...

zsalch: The isPondering is false, which is the condition ` && !isPondering` that when playing with AI, the ponder is turned off. The `togglePonder();` just activates pondering, which must be reset to isPondering = false again, because the current user sees an inactive ponder status.

OlivierBlanvillain:
LGTM

Issue #351, opened 2018-08-23
Feature-request: Blunder check mode


killerducky:
I would like to have ponder, current winrate, last move +/-, and eval graph all on. But turn off the board display of winrates and variations. This way I can play out moves/variations live and notice when LZ says I made a mistake, but I won't know what the correct move is. I can undo and try again until I find what LZ thinks is a better move.

fsh:
I've sorely wanted something like this too, had a hack for forcing it to turn off variations on board, but it required a different copy and wasn't a very good solution.

But actually found a project based on Lizzie has this very thing: https://github.com/aerisnju/mylizzie

They don't have any releases so you'll need to compile it yourself, but it's pretty great. You can turn off on-board move suggestions with a keyboard shortcut and still keep doing variations etc., having it either list the moves that were less than so-and-so many negative diff from last move in value score, or when you notice the graph plummets.

AncalagonX:
I would prefer a display like Crazy-Sensei has, which is a bar chart with one bar per move showing how much the winrate changed on that move.

Example:

![2018-10-01_20-34-25_ chrome](https://user-images.githubusercontent.com/37827291/46286339-cdf7f080-c5b9-11e8-927e-e6f18fad47c2.png)

The blue line shows winrate, with a small middle divider line at 50% tie winrate. The red bars are what I am talking about.

A red bar *above* the middle divider line means white made a mistake that moved the winrate higher in favor of black. Likewise, a red bar *below* the middle divider line is a black mistake, which moved the winrate in favor of white. The size of the red bar indicates the degree of winrate change (i.e. how big of a blunder it was).

This has always been the best implementation/visualization for "blunder detection" I've seen to date. I find it to be much more immediately informative than just looking at the simple winrate line graph.

AncalagonX:
Here's move 137 from that game above, the biggest "blunder" of the kifu. It was a game against LZ from many, many, many moons ago:

![2018-10-01_20-43-04_ chrome](https://user-images.githubusercontent.com/37827291/46286776-54610200-c5bb-11e8-9ab8-62fbf202a175.png)

The blue square on the board is the blunder, where black (LZ) played next. The [1] [2] [3] on the board are the best three sequences that Crazy-Sensei found. [Full game here on Crazy-Sensei.](https://www.crazy-sensei.com/?lang=en&location=game&game=845&move=1)

I'd be super happy for a long time just having the "red bar graph" implemented and nothing else.

featurecat:
[...]

PR #384, opened 2018-10-05
next: Display blunder bar in the winrate graph

zsalch:
#351 

I'm trying to display blunder bar in the winrate graph.

* 1. Add new option 'show-winrate-blunder-bar' in the config.txt to turn on blunder bar, default is turn off.
* 2. Can specify the minimum width of the blunder bar by the option 'minimum-winrate-blunder-bar-width' in the config.txt, default is 3.
* 3. Considering that the winrate change should be mostly below 50%, an option is added to weighted display the height of the bar, that is, the change below 50% increases the display height, and the above 50% reduces the height. Currently only a simple weighted, the winrate change below 50% uses 75% of the display area, the above uses 25% of the display area. The default is off

turn off -> turn on
image
Adjusted for the green of the winrate graph should always be on top:
image

* 4. Due to the small width of the winrate graph, when the move number is large, the display is crowded, so an option 'dynamic-winrate-graph-width' is added to dynamically adjust the width, so that can see clearly when the number of steps is small. Default is turn off.

* 5. Add new option 'winrate-stroke-width' in the config.txt to set the winrate line width, default is 3.

* 6. Add large winrate mode, press Ctrl+W to toggle:
lb

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
+++ b/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
@@ -137,7 +137,7 @@ public class LizzieFrame extends JFrame {
     variationTree = new VariationTree();
     winrateGraph = new WinrateGraph();
 
-    setMinimumSize(new Dimension(640, 480));
+    setMinimumSize(new Dimension(640, 400));
     JSONArray windowSize = Lizzie.config.uiConfig.getJSONArray("window-size");
     setSize(windowSize.getInt(0), windowSize.getInt(1));
     setLocationRelativeTo(null); // Start centered, needs to be called *after* setSize...
@@ -189,6 +189,13 @@ public class LizzieFrame extends JFrame {
             Lizzie.shutdown();
           }
         });
+  }
+
+  /** Clears related status from empty board. */
+  public void clear() {
+    if (winrateGraph != null) {
+      winrateGraph.clear();
+    }
   }
 
   public static void startNewGame() {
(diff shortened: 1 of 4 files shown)


featurecat:
thanks. I think the green of the winrate graph should always be on top. is it possible to make it more visible? it's confusing in the middle of your graphs.

featurecat:
also I absolutely love the name "blunder bar" 😁

zsalch:
I'm confuse too.
Need a solution to set the drawline and fillrect z-order.

zsalch:
Adjusted:
image

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -334,117 +352,286 @@ public void paint(Graphics g0) {
...
+          vh = panelH;
+          // subboard
+          subBoardY = topInset;
+          subBoardWidth = panelW - leftInset;
+          subBoardHeight = panelH;
+          subBoardLength = Math.min(subBoardWidth, subBoardHeight);
+          subBoardX = statx + (vw - subBoardLength) / 2;
+        }
+      } else {
+        // Portrait mode
```
OlivierBlanvillain: It looks like some code from #398 leaked into this PR. Could you try to clean this up before the review?

zsalch: It not as a portrait mode, but just a display improvement, which reduces the user's error when resizing the window.

featurecat:
Should I include this in Lizzie 0.6, and should I wait until you're done reviewing to do so?

PR #385, opened 2018-10-06
New board and background

ParmuzinAlexander:


zsalch:
@featurecat @OlivierBlanvillain 
I think that in order to demo and guide users to use the new Theme, we should include an optional Theme when releasing the new version. What do you think?
@ParmuzinAlexander 
If so, could you like to provide an extra background, board and stone images that you like?

OlivierBlanvillain:
@zsalch Sounds good! Should these additional images be added on top of the your theme PR?

zsalch:
@ParmuzinAlexander 
Thanks. I will add to my theme PR.
But I don't know what the name is, can you give them a proper name?

OlivierBlanvillain:
@zsalch Maybe you could also include in the sabaki images as well (with proper attribution)? As long as it's not the default, I think these would be good to have!

ParmuzinAlexander:
@zsalch 
Theme name? If use one folder for everything images it will be easier to customization. Complete theme like "sabaki", "photorealistic" this is a step backwards.
Now "assets" and "theme" they have the same functionality. It would be logical remove "assets" and use only "theme" folder (contains all the pictures.).  And theme.txt transfer inside "ui" config.txt.

zsalch:
@ParmuzinAlexander 
I mean, can you name your theme that includes you provided the images?

ParmuzinAlexander:
@zsalch 
Yasnaya

zsalch:
@ParmuzinAlexander 
I guess what you mean is that all the image files are placed in the theme folder, and the corresponding image is set in the theme.txt, such as yasnaya.txt, and the corresponding theme is specified as 'yasnaya ' in the config.txt file.
I want to know what is the advantage of this approach over the use of separate folder. I guess is it easy to maintain image files?

ParmuzinAlexander:
@zsalch 
If user does not like some detail example background, it can change 1 character in config.txt  background3.png->background1.png without copying and replacing the real file. All files weigh only 1.4 Mb.
And I don't suggest using theme.txt or yasnaya.txt need just setting what texture to use in config.txt.

OlivierBlanvillain:
I agree with you @ParmuzinAlexander, I feel that only having to update `config.txt` makes things simpler...

zsalch:
The new Theme logic supports two ways for change option.
You can ignore the theme.txt and directly change option in the config.txt.

OlivierBlanvillain:
Let's merge this for now and continue discussion on #378. @ParmuzinAlexander Thanks!

PR #386, opened 2018-10-06
Setup automatic code formatter using the Google Java Style

OlivierBlanvillain:
`mvn compile` now automatically reformats the entire codebase. They are also pugins available for Eclipse & IntelliJ, see https://github.com/google/google-java-format.

This PR is best reviewed commit by commit given that the 2nd one is an auto generated diff. The 3rd commit does some very light refactoring to make the code fit nicely in the 100 line width limit enforced by the formatter.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,18 @@
                     <useFile>false</useFile>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>com.coveo</groupId>
+                <artifactId>fmt-maven-plugin</artifactId>
+                <version>2.5.1</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>format</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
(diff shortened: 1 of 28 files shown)


OlivierBlanvillain:
@featurecat @zsalch What do you think?

featurecat:
useful, how does it get set up? whenever programmers compile the program, it runs?

featurecat:
so cool!! I'll check it in then later today I'll start drafting a release for lizzie 0.6. any concerns? also I don't know why the PR for new texture was closed but I'm just going to merge the new textures from there.

OlivierBlanvillain:
@featurecat I think we've been merging things PR a bit too eagerly, for instance here I introduced a couple of bugs (such as using `x` twice here https://github.com/featurecat/lizzie/pull/386/commits/43153836698eb0b3347a903e66dba533c5ce12d2#diff-c51e6645e657a3c7c444daafb9ad21e0R941, one of the things that @zsalch fixed in #393...). I propose that from now on we don't merge anything until it has been properly reviewed and received a "LGTM" comment indicating that the reviewer is happy with the diff. What do you think?

zsalch:
@OlivierBlanvillain 
The automatically reformat is very good! I am already using it.
Recommended as a code standard for contributors to keep the code in the same style.
At the same time, we can also share a code style configuration file, such as [Google Style Guides](https://github.com/google/styleguide).

featurecat:
sounds good sorry for checking in too quickly. I'll let you check in more things in the short term.

Issue #268, opened 2018-06-01
Automatic Network Downloading temporarily broken...


featurecat:
It looks like the page which I used for network detection (http://zero.sjeng.org/network-profiles) changed its syntax just as we released Lizzie 0.5. So currently, Lizzie cannot decipher the new format, so it can't detect when it needs to download a new network. This will be fixed for the next version of Lizzie...

WYVYW:
I doubt that they did this on purpose, but I would be great if we could support Leela more.
possible ways:
- including autogtp while using lizzie as sort of screensaver instead of leela-watcher
- asking to start autogtp (in background) when closing lizzie
- or sort of nag-screen after closing (until autogtp is implemented)

roy7:
Whoops, could you post an issue on server github if something like this happens again? I don't think we had any idea you were using network-profiles for that. 

Is there a way to just check things like timestamp and file size of best-network.gz for changes?

You can also use the http://zero.sjeng.org/best-network-hash endpoint to get the official best network hash at any time. If it doesn't match the hash of the best network you're currently using, then you know to download best-network.gz again.

featurecat:
oh I didn't know we had a hash url. that's exactly what I wanted. thanks!

roy7:
Cool then we're set. :) The network-profiles and various other web interface are all under construction atm and not 100% safe to use for unintended purposes. :)

PR #387, opened 2018-10-06
Remove network downloading feature

OlivierBlanvillain:
Based on #386, only the last commit is new.

Downloading the latest network is currently broken, see #268.

In my opinion it makes little sense to have this built in into lizzie, it's much simpler to explain to users that they can download anything from http://zero.sjeng.org/ to replace the `network.gz` that we ship with the release, and they should be good to go!


src/main/resources/l10n/DisplayStrings.properties
```diff
@@ -47,11 +47,10 @@ LizzieFrame.prompt.sgfExists=The SGF file already exists, do you want to replace
 LizzieFrame.prompt.showControlsHint=hold x = view controls
 LizzieFrame.prompt.switching=switching...
 LizzieFrame.display.lastMove=Last move
-LizzieFrame.display.pondering=Pondering 
+LizzieFrame.display.pondering=Pondering
```
zsalch: If a space is removed here, then a space is added in the LizzieFrame for draw the pondering text.

OlivierBlanvillain: Good catch, fixed in the latest commit!

OlivierBlanvillain:
@zsalch Are you done with the review? (Please write a LGTM when you are :smile: )

src/main/java/featurecat/lizzie/Util.java
```diff
@@ -1,110 +0,0 @@
-package featurecat.lizzie;
```
zsalch: I hope to keep Util.java for some general functions.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -439,13 +441,13 @@ public void paint(Graphics g0) {
...
-                  + resourceBundle.getString(pondKey);
-          String switchText = resourceBundle.getString("LizzieFrame.prompt.switching");
-          String weightText = Lizzie.leelaz.currentWeight();
-          String text = pondText + " " + weightText + (Lizzie.leelaz.switching() ? switchText : "");
+          String statusKey = "LizzieFrame.display." + (Lizzie.leelaz.isPondering() ? "on" : "off");
+          String statusText = resourceBundle.getString(statusKey);
+          String ponderingText = resourceBundle.getString("LizzieFrame.display.pondering");
+          String switching = resourceBundle.getString("LizzieFrame.prompt.switching");
+          String switchingText = Lizzie.leelaz.switching() ? switching : "";
+          String weightText = Lizzie.leelaz.currentWeight().toString();
```
zsalch: The currentWeight() is a String. Why need toString here?

src/main/java/featurecat/lizzie/Util.java
zsalch: It's ok.

zsalch:
LGTM.

Issue #383, opened 2018-10-04
Use different color to indicate move blunder in variation window. 


phiming:
Use different color to indicate move blunder in variation window. 
This also partly resolve the issue  #351 

![image](https://user-images.githubusercontent.com/35087044/46484777-03f0db00-c82d-11e8-8252-6671ca30b3a9.png)

In the right variation windows, you will see some colorful node, which means:
Red,  serious blunder
Yellow, questionable blunder
Green, good move
grey, normal move
....
What color is the best, and how much win rate drop can be serious?  I need some suggestion.

Win rate comes from history record, but if there is real time statistics from leelazero engine, it will be used first.

phiming:
Nobody has suggestions?
So, I will give my suggestion, and follow it.
winrate drops 
-       < %2, normal, no special color
-       %2 - %6,  questionable, yellow
-       6%- 10%,  clear mistake, orange 
-       %10-30%, serious blunder, red,
-       >%30, deadly, black

concentric circles suggest this node has comments

zsalch:
It may be better to allow users to set these thresholds

phiming:
![21](https://user-images.githubusercontent.com/35087044/46571685-f877ee00-c9ab-11e8-869c-e873ca123a03.jpg)

Concentric circles suggest this node has comments

phiming:
@zsalch 
Good suggestion, maybe they can use config file to set those thresholds.

PR #389, opened 2018-10-06
Use different color to indicate move blunder in variation window

phiming:
#383
Use different color to indicate move blunder in variation window.
Add concentric circles display for nodes with comments.


zsalch:
@phiming 

@OlivierBlanvillain has added a auto formatter plug-in into the pom.xml in latest branch 'next'.
It is recommended to commit the formatted code again.

OlivierBlanvillain:
@phiming It's a nice idea, but I think it makes more sense to display this information alongside the winrate. Indeed, at any point in time we can only see a small subset of the nodes in the variation window, whereas the winrate graph shows an overall view of the game. So my suggestion would be to close this in favor of #384.

featurecat:
no I like this idea. why not display it?

featurecat:
can I have an explanation of the colors meanings? I don't get why there is black

phiming:
@zsalch
It seems this function will not be merged very soon, so if you need it right now, you may get it from my homepage  Please let me know if any question.

phiming:
@Olivier
Sorry for I use phone to type, cant type too much.
some people like complicated thing and want to master everthing, some like simple and easy thing.  though we cant see all blunders in variation window, it should be enough for some one. so, why not to keep it. frankly, i think its a little crowd and complicated to dipslay all blunders in winrate window.

phiming:
@Olivier
Sorry for I use phone to type, cant type too much.
some people like complicated thing and want to master everthing, some like simple and easy thing.  though we cant see all blunders in variation window, it should be enough for some one. so, why not to keep it. frankly, i think its a little crowd and complicated to dipslay all blunders in winrate window.

phiming:
@featurecat
I opened an issue to get some suggestion.
currently, black means deadly blunder. its not final decision, any suggestion will be considered.

featurecat:
what about red orange yellow? you could even make it a smoother gradient.

zsalch:
@phiming 
I think this feature is still good, There are a few things to improve:

1. Now the 'next' branch has changed a lot, so the commits are too different now. It is recommended to rebase the current commits and only submit the difference from the latest next;
2. as the comment of , the node display is not smooth enough, of course, the current version may also have this problem, will better if it can be improved;
3. Thresholds for different colors can be configured by the user, such as:
     "blunder-thresholds": [2, 6, 10, 30]
     If this PR is approved, then can consider adding color settings to the theme in the future.
4. The node with comments feels that the color is more conspicuous than the concentric circle.

OlivierBlanvillain:
I'll take care 1.; merging next into this branch.

OlivierBlanvillain:
By the way, why is this PR based on the "Use mouse to jump to any node(position) in variation window"? If it doesn't have to it would be better to make independent commits, because as it stands we will have to wait for:

- All the comments in #391 to addressed
- #391 to be merged
- This PR to be rebased on top of next

Before even starting to review the code...

phiming:
@zsalch, for
1, I will rebase it.  Though everytime I was painful with so many conflicts. It ruined funs of programming.

2, 4, I am not good at UI designing, in fact, this feature needs only a little coding,  but the ideas  is nice, at least for me.  You are more experienced at UI designing, and also have good program skill.  You may change the code as you like, I would be very glad for your help.

OlivierBlanvillain:
@phiming You can also merge if rebasing is too painful (and it probably is here).

zsalch:
@phiming 
I'll try to make it better.

zsalch:
@OlivierBlanvillain @phiming 
I have tried to do:
image

Config:
"blunder-winrate-thresholds": [-30,-20,-10,-5,5,10],
"blunder-node-colors":[[255,0,0],[0,255,0],[0,0,255],[255,255,0],[0,255,255],[255,0,255]],

What do you think, what suggest? I always feel that the current node is a little bit strange.

zsalch:
@phiming @OlivierBlanvillain @featurecat 

With comment node:
image

This effect does not look good.
Perhaps using a color marker blunder is not a particularly good choice because the current node and the node with comments have to use other schemes.
Anybody have any suggestions?

featurecat:
I think this effect looks good to indicate comments. for blunders did you try experimenting with less saturated colors? something like that?

zsalch:
Allow users to specify the blunder color, like [255,0,0] or [255,0,0,200].

zsalch:
Reopen #413, closed this.

PR #393, opened 2018-10-08
Fixed some merge issues

zsalch:
Fixed some merge issues.

### DIFF
--- a/src/main/java/featurecat/lizzie/analysis/Leelaz.java
+++ b/src/main/java/featurecat/lizzie/analysis/Leelaz.java
@@ -64,6 +64,7 @@ public class Leelaz {
   private String engineCommand = null;
   private List<String> commands = null;
   private JSONObject config = null;
+  private String currentWeightFile = null;
   private String currentWeight = null;
   private boolean switching = false;
   private int currentEngineN = -1;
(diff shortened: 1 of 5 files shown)


OlivierBlanvillain:
@zsalch LGTM

src/main/java/featurecat/lizzie/analysis/Leelaz.java
```diff
@@ -114,10 +115,6 @@ public void startEngine(String engineCommand) throws IOException {
       return;
     }
 
-    File startfolder = new File(config.optString("engine-start-location", "."));
-    String networkFile = config.getString("network-file");
-    // substitute in the weights file
-    engineCommand = engineCommand.replaceAll("%network-file", networkFile);
```
infinity0: Why was this removed?

infinity0:
Can your code review standards be a bit stronger? "Fixed some merge issues" is a bad PR description, and LGTM in response to that is not doing proper code review.

src/main/java/featurecat/lizzie/analysis/Leelaz.java
infinity0: Specifically I mean the `%network-file` part. I can see the other parts were moved elsewhere so those seem fine.

OlivierBlanvillain: Seems redundant with what's done in the constructor.

OlivierBlanvillain:
@infinity0 Yeah, we're getting there...

For context this project got to this state without any code review, tests or CI. This PR fixes stuff I introduced in 43153836698eb0b3347a903e66dba533c5ce12d2 which was accidentally merge without being review (except for what you pointed out, which I overlooked). If you want to give us a hand feel to review whatever is in the PR queue!

PR #394, opened 2018-10-08
Setup Travis CI

OlivierBlanvillain:
This PR setups Travic CI to check the following on every pull request:

- The code compiles
- The test pass
- The code conforms to the Google Java Style (which everyone should get for free when running `mvn compile`)

I've included two shell scripts to compile Leela Zero and download a latest 192x15 network, both of which are required to get the tests passing. I've also setup the Travis caches to prevent recompiling leelaz and re-dowloading the network on every PR. As a result, the CI is blasting fast: it runs under [45 seconds](https://travis-ci.org/OlivierBlanvillain/lizzie/builds/438407131?utm_source=github_status&utm_medium=notification) (but that's also because we have like two tests ^^').

I had to use the Mac OS image from Travis because the Linux they provide is too old to compile leelaz, and switching to mac/brew was simpler than recompiling boosts from source on the CI. I left the `apt` commands to install the leelaz dependencies on Debian based distro for documentation and possibly future uses once Travis updates their Linux distribution.

@featurecat To finalize this I now need to go to https://travis-ci.org/, sign in, and unable the `featurecat/lizzie`. It shouldn't take more than 5 minutes, and after you've done that and we merged this PR we should get every pull request compiled and tested from Travis!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: java
+jdk: oraclejdk8
+
+# The latest linux image on Travis is too old to build leelaz, what a shame...
+os: osx
+
+cache:
+  directories:
+  - $HOME/.m2
+  - $HOME/.leelaz
+
+install:
+  - sh scripts/make-leelaz.sh
+  - sh scripts/download-network.sh
+
+script:
+  - mvn test
+  - mvn com.coveo:fmt-maven-plugin:check
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ Lizzie is a graphical interface allowing the user to analyze games in
 real time using [Leela Zero](https://github.com/gcp/leela-zero). You
 need Java 8 or higher to run this program.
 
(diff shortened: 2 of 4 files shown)


OlivierBlanvillain:
@zsalch What do you think?

zsalch:
I like it, I think it will help us validate & format the submitted code.

featurecat:
Yes, I'll handle this soon. Thanks

OlivierBlanvillain:
@featurecat It's been a month, we still need your input to activate the CI. Please have a look, it's probably doable from a phone...

featurecat:
right, sorry. I'll do it after classes today.

featurecat:
Ok I've done it, what next?

OlivierBlanvillain:
That's it, now every commit pushed on a PR will be automatically compiled and tested by Travis, and we get a nice error reporting in the UI :smile:

Issue #395, opened 2018-10-09
Deals deals badly with portrait mode resolutions


fsh:
Starting current next branch on a monitor in portrait mode (rotated 90 degrees) gives the following exception in a loop:

```
Exception in thread "AWT-EventQueue-1" java.lang.IllegalArgumentException: Width (0) and height (32) cannot be <= 0
	at java.awt.image.DirectColorModel.createCompatibleWritableRaster(DirectColorModel.java:1016)
	at java.awt.image.BufferedImage.<init>(BufferedImage.java:333)
	at featurecat.lizzie.gui.LizzieFrame.drawPonderingState(LizzieFrame.java:571)
	at featurecat.lizzie.gui.LizzieFrame.paint(LizzieFrame.java:449)
	at javax.swing.RepaintManager$4.run(RepaintManager.java:842)
	at javax.swing.RepaintManager$4.run(RepaintManager.java:814)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at javax.swing.RepaintManager.paintDirtyRegions(RepaintManager.java:814)
	at javax.swing.RepaintManager.paintDirtyRegions(RepaintManager.java:789)
	at javax.swing.RepaintManager.prePaintDirtyRegions(RepaintManager.java:738)
	at javax.swing.RepaintManager.access$1200(RepaintManager.java:64)
	at javax.swing.RepaintManager$ProcessingRunnable.run(RepaintManager.java:1732)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at org.GNOME.Accessibility.AtkWrapper$6.dispatchEvent(AtkWrapper.java:715)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
[...]

PR #398, opened 2018-10-11
next: Fixed display issues

zsalch:
Issue #395 
Base on #378 #384 

It can support portrait mode, although it is not perfect, but it works.
In the portrait mode, the layout remains to be discussed, but I think it can help the current display.
Also fixed some display issues, when the user feels free to change the window size.

pm1
pm2
pm3

### DIFF
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -1,5 +1,7 @@
 package featurecat.lizzie;
 
+import featurecat.lizzie.theme.Theme;
+import java.awt.Color;
 import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


AncalagonX:
If possible, it would be nice to make this a user-configurable setting to have it display above *or* below the board.

OlivierBlanvillain:
I'm not really a fan of this solution as is doesn't fully solves the problem: people want to customize the layout. What about using something like [JToolBar](https://docs.oracle.com/javase/tutorial/uiswing/components/toolbar.html)? It seems like think can be moved around inside the main windows or even detached.

zsalch:
I agree that this is just a transitional solution.
The best solution is that each panel can be customized, blended or separated.
It seems that JToolBar can implement, which may be the next goal.

OlivierBlanvillain:
> I agree that this is just a transitional solution.

Then I would be for postponing this PR until we have a definitive solution...

zsalch:
I no longer use this as a portrait mode solution, but just to solve some display issue, so I have changed the title.
Or I should close this, then wait for #402 to merge and re-open a PR that fixes the display issues.

PR #400, opened 2018-10-13
Fix #397 (broken autosave)

### LINKED ISSUE #397: Lizzie does not autosave because "persist" never gets "autosave" added to it
Using Lizzie /next/ built today (and any built ever since the original autosave PR #298), autosaving is broken because the "autosave" entry is always missing from the `persist` file. Deleting `persist` and having Lizzie regenerate from scratch also never adds the "autosave" entry.

Lizzie should add the "autosave" line to the `persist` file automatically, but it doesn't right now.

Here is an example of a working `persist` file that autosaves to the persist file:

```
{
  "autosave": "",
  "ui-persist": {},
  "filesystem": {"last-folder": "C:\\Lizzie_SGFs"}
}
```

It seems all autosave functions work properly though (autosave time interval, auto-reloading board state, board change detection, etc.), as long as we add the above "autosave" line to the `persist` file. The solution for making this happen is probably straightforward for those who know Java better than me.

kaorahi:


OlivierBlanvillain:
How does one turn autosave off after this change?

kaorahi:
One can set

    "resume-previous-game": false,

or

    "autosave-interval-seconds": -1,

in config.txt (#298).

OlivierBlanvillain:
Thanks!

Issue #301, opened 2018-06-17
Wrong move number color over dead stones


ParmuzinAlexander:
Numbers 10 and 12 on black stones, but shows white dead stones. https://imgur.com/6bEfK1G

featurecat:
I just didn't want to get around to fixing this haha. It's pretty minor... I would fix it eventually but I'm not very concerned about it.

OlivierBlanvillain:
@ParmuzinAlexander Could you upload a sgf file?

ParmuzinAlexander:
@OlivierBlanvillain 
`(;GM[1]FF[4]SZ[19]HA[0]KM[7.5]CA[UTF-8]PB[]
BR[]PW[]WR[]DT[]RE[];B[qd];W[dd];B[pq];W[dp];
B[fq];W[qo];B[op];W[oc];B[cn];W[pe];B[qe];W[pf];B[qg];W[dn];B[dm];W[en];
B[cp];W[cm];B[co];W[dl];B[dq];W[em];B[fc];W[fd];B[gd];W[fe];B[dc];W[cc];
B[ec];W[cd];B[jc];W[gi];B[cf];W[ch];B[dh];W[ci];B[gg];W[hf];B[hd];W[gf];
B[di];W[cj];B[pc];W[pb];B[qb];W[oh];B[gk];W[fj];B[ih];W[hg];B[qp];W[po];
B[oo];W[pm];B[om];W[ol];B[nm];W[qh];B[qf];W[kc];B[kd];W[ld];B[lc];W[kb];
B[lb];W[le];B[mc];W[pd];B[qc];W[jd];B[ke];W[kf];B[je];W[ie];B[jf];W[id];
B[jb];W[jg])`

zsalch:
@ParmuzinAlexander 
The sgf seems not the same as mentioned in the issue.

ParmuzinAlexander:
@zsalch 
Same. But actually there is a more global problem double move prediction in one place not shows. But how to solve I don't know.

zsalch:
@ParmuzinAlexander 
I understand what you mean, I will provide a test version later to determine if this problem can be solved.

zsalch:
I have a question: does need remove the dead stones when display the ghost branch?

zsalch:
@ParmuzinAlexander 
There is a beta version here: [pre-release](https://github.com/zsalch/lizzie/releases/tag/v0.5%2B)

ParmuzinAlexander:
[...]

PR #401, opened 2018-10-14
fixed: Display latest stone for ghost dead stone

zsalch:
Issue: #301 

Display latest stone for ghost dead stone


src/main/java/featurecat/lizzie/gui/BoardRenderer.java
```diff
@@ -384,7 +384,8 @@ private void drawBranch() {
 
     for (int i = 0; i < Board.BOARD_SIZE; i++) {
       for (int j = 0; j < Board.BOARD_SIZE; j++) {
-        if (Lizzie.board.getData().stones[Board.getIndex(i, j)] != Stone.EMPTY) continue;
+        // Display latest stone for ghost dead stone
+        // if (Lizzie.board.getData().stones[Board.getIndex(i, j)] != Stone.EMPTY) continue;
```
OlivierBlanvillain: I would just remove the line and don't put a comment...

zsalch: Currently some test is ok, but has not more tests. I want to temporarily reserve it so that can suggest changes here when there is a problem.
Of course, I'm ok with remove it.

OlivierBlanvillain: I see. So you propose to keep it commented out until we are sure that it doesn't break something else?

zsalch: Yes, this change can solve some problems, but it also changes the display of the stones. I think we should be cautious.

zsalch: I have committed a new change for fix the stone double shadow issue when branch

OlivierBlanvillain:
LGTM

PR #403, opened 2018-10-18
Fix #396? Or at least improve parsing of LZ ponder output

### LINKED ISSUE #396: no variation moves show when pondering is on
Hello,
i have downloaded lizzie for linux and made leelaz from the next branch of leela-zero.
When pondering is on, I can see the proposed moves but the variation (sequence of moves) does not show neither on the main board nor on the small board on the left.
I get this :
 Exception in thread "AWT-EventQueue-0" java.lang.NumberFormatException: For input string: ""

Regards

sinf:
The latest Leela Zero (0.15) outputs an additional "prior"-field, breaking the code that parses LZ's output.
I changed the code more robust against changes in LZ output.
Java >=7 has a string switch but I like string.equals for some reason.
This change probably fixes #396 atleast for me it did

### DIFF
--- a/src/main/java/featurecat/lizzie/analysis/MoveData.java
+++ b/src/main/java/featurecat/lizzie/analysis/MoveData.java
@@ -19,16 +19,27 @@ public class MoveData implements Comparable<MoveData> {
      * @param line line of ponder output
      */
     public MoveData(String line) throws ArrayIndexOutOfBoundsException {
+
         String[] data = line.trim().split(" ");
 
-        // Todo: Proper tag parsing in case gtp protocol is extended(?)/changed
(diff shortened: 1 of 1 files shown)


OlivierBlanvillain:
@sinf I looks like you made a commit on top master, which is 5 month old, all the recent development is happening on `next`. Could you check if the issue you are fixing is not already fixed on `next`?

sinf:
> @sinf I looks like you made a commit on top master, which is 5 month old, all the recent development is happening on `next`. Could you check if the issue you are fixing is not already fixed on `next`?

Damn. It is fixed already. I had leela zeros next-branch but lizzies master.
Closing this PR then.

PR #413, opened 2018-10-31
next: Support use different color to indicate move blunder in variation

zsalch:
Base on #412 
Closes #389 

Color node:
image

With comment node:
image

1. Config:
"blunder-winrate-thresholds": [-30,-20,-10,-5,5,10],
"blunder-node-colors":[[255,0,0],[0,255,0],[0,0,255],[255,255,0],[0,255,255],[255,0,255]],
"comment-node-color":[0,0,255,255],

2. Ctrl+T for toggle the node color with the comment

3. Press ‘Y' for switch node color display

   default (show color) -> only black color -> only white color


featurecat:
LGTM

featurecat:
By the way, you forgot to add the three config values to the config file. So I had to add them to get it to work. I fixed it. You should try deleting the config file as a test before making a PR, to see if a new feature still works.

featurecat:
Actually apparently the node color is part of the theme. I'm not sure how I got it to work but it no longer works as of now. It's not part of the default theme which is problematic.

Issue #28, opened 2018-03-27
can lizzie offer/adopt the waltheri/wgo.js theme/color/style 


hydrogenpi:
I like the look and style of this better

https://github.com/waltheri/wgo.js

Can it be made an optional theme?

featurecat:
Yes, the next version will have moddable themes.

featurecat:
You can now swap out the files in the Assets folder for your own. If you do create this theme, please share it here for other users! Thanks :)

PR #414, opened 2018-10-31
next: Support click the node in the variation

zsalch:
Base on #412 #413 
Closes #391 

Support click the node in the variation


featurecat:
The travis build failed -- could you take a look?

zsalch:
I committed again, but the travis check seems auto cancelled.

> This

 job was cancelled because the "Auto Cancellation" feature is currently enabled, and a more recent build (#28) for pull request #414 came in while this job was waiting to be processed.

featurecat:
Oh interesting. I'm not sure if I like that feature or not. Restarted the build.

PR #415, opened 2018-10-31
next: Add more keyboard shortcut

zsalch:
Add more keyboard shortcut:

1. Press key 'M' for switch the move number display

   Not show -> Show all move number -> Show last move number

2. Press key Ctrl + Left arrow to jump the variation top

3. Press key Ctrl + Shift + Up/Down to jump the previous/next the node with comment

4. Press Ctrl+Home to clear board


src/main/java/featurecat/lizzie/rules/BoardHistoryNode.java
```diff
@@ -322,4 +322,56 @@ public boolean isMainTrunk() {
...
+    int moves = 0;
+    while (node.next().isPresent()) {
+      moves++;
+      BoardHistoryNode next = node.next().get();
+      if (!next.getData().comment.isEmpty()) {
+        break;
+      }
+      node = next;
+    }
+    return moves;
```
OlivierBlanvillain: The int returned here is never used, same comment for the next method.

src/main/java/featurecat/lizzie/rules/BoardHistoryNode.java
```diff
@@ -322,4 +322,56 @@ public boolean isMainTrunk() {
...
+      BoardHistoryNode previous = node.previous().get();
+      if (!previous.getData().comment.isEmpty()) {
+        return Optional.ofNullable(previous);
+      }
+      node = previous;
+    }
+    return Optional.empty();
+  }
+
+  public int getNextNodeWithCommentMoves() {
```
OlivierBlanvillain: The get in the name is misleading. This mutates the board by moving to a new board. Therefor the name should be `goToNextNodeWithComment` instead. Same comment for the next method.

src/main/java/featurecat/lizzie/rules/BoardHistoryNode.java
```diff
@@ -322,4 +322,56 @@ public boolean isMainTrunk() {
     }
     return true;
   }
+
+  public Optional<BoardHistoryNode> findNextNodeWithComment() {
```
OlivierBlanvillain: The method and the next one are never used.

src/main/java/featurecat/lizzie/gui/Input.java
```diff
@@ -66,6 +66,20 @@ private void undoToChildOfPreviousWithVariation() {
     if (!Lizzie.board.undoToChildOfPreviousWithVariation()) Lizzie.board.previousMove();
   }
 
+  private void undoToFirstParentWithVariations() {
+    if (Lizzie.board.undoToChildOfPreviousWithVariation()) {
+      Lizzie.board.previousMove();
+    }
+  }
+
+  private void goCommentNode(int direction) {
```
OlivierBlanvillain: Why use an int if you only expect two values? I would use a boolean instead, called `moveForward`.

src/main/java/featurecat/lizzie/Config.java
```diff
@@ -198,7 +200,16 @@ public boolean mergeDefaults(JSONObject config, JSONObject defaultsConfig) {
   }
 
   public void toggleShowMoveNumber() {
-    this.showMoveNumber = !this.showMoveNumber;
+    if (this.showMoveNumber && uiConfig.has("only-last-move-number")) {
+      if (this.onlyLastMoveNumber == 9999) {
```
OlivierBlanvillain: I would use `-1` as a marker value, and put that as a `private static final int` field of the Config class. I think it would look better than the current solution where this constant is duplicated all over the code...

src/main/java/featurecat/lizzie/rules/BoardHistoryNode.java
zsalch: It return value will be used by the function goCommentNode in the Input.java

src/main/java/featurecat/lizzie/rules/BoardHistoryNode.java
zsalch: Changed.

src/main/java/featurecat/lizzie/rules/BoardHistoryNode.java
zsalch: Removed.

src/main/java/featurecat/lizzie/gui/Input.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/Config.java
zsalch: Fixed.

src/main/java/featurecat/lizzie/rules/BoardHistoryNode.java
OlivierBlanvillain: Alright, then you might wan't to add a one line javadoc `@returns ...` :)

PR #416, opened 2018-10-31
next: Support auto detect the gib/sgf file encoding

zsalch:
Support auto detect the gib/sgf file encoding using lib 'juniversalchardet'.


src/main/java/featurecat/lizzie/rules/GIBParser.java
```diff
@@ -20,8 +21,9 @@ public static boolean load(String filename) throws IOException {
       return false;
     }
 
+    String encoding = EncodingDetector.detect(filename);
```
OlivierBlanvillain: Why not use `UniversalDetector.detectCharset` directly?

src/main/java/featurecat/lizzie/util/EncodingDetector.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import org.mozilla.universalchardet.UniversalDetector;
+
+public class EncodingDetector {
+
+  public static String detect(String fileName) {
+    String encoding = "UTF-8";
+    try {
+      encoding = detect(new FileInputStream(fileName));
```
OlivierBlanvillain: Why not use `UniversalDetector.detectCharset` directly?

zsalch: I can't find the function and by example of the official site.

PR #420, opened 2018-11-03
next: New theme yasnaya by ParmuzinAlexander

zsalch:
New theme 'yasnaya' by @ParmuzinAlexander
As discussion: #385 

Because the 'board1.png' already as the default board, so use the 'board7.png':

image

1. Allow user turn on/off the border by option 'show-border', default is turn off.

2. And new theme options are added:

  "winrate-stroke-width": 2,
  "minimum-blunder-bar-width": 3,
  "comment-background-color": [0, 0, 0, 200],
  "comment-font-color": [255, 255, 255, 200],
  "comment-node-color": [0, 0, 255, 200],
  "winrate-line-color": [0, 255, 0, 200],
  "winrate-miss-line-color": [0, 0, 255, 200],
  "blunder-bar-color": [255, 0, 0, 200],
  "blunder-winrate-thresholds":[-30,-20,-10,-5,5,10],
  "blunder-node-colors":[[255,0,0,200],[0,255,0,200],[0,0,255],[255,255,0],[0,255,255],[255,0,255]]

3. and font setup:
    "ui-font-name": "Wawati SC",
    "winrate-font-name": "Wawati SC Bold",
    "font-name": "Wawati SC",

PS: above the options are effective for both config.txt & theme.txt

@ParmuzinAlexander
About these new options, do you have some better suggestion?


ParmuzinAlexander:
@zsalch 
Everything is fine except the font in the screenshot. But I see normal font with "Wawati SC". Font should be clear and understandable.

zsalch:
The wawati SC just a test font, current the theme 'yasnaya' does not setup a font, it will use default font.

zsalch:
The board border seems to have been there, not this change.
No board border, do you mean like following? Maybe we should allow the user to remove all the borders, because now almost all panels have it.
image

ParmuzinAlexander:
@zsalch 
It's perfect!

ParmuzinAlexander:
@zsalch 
I replaced board7 to the original texture. But it is not very important, visually they are the same.

featurecat:
I added a default config option: `theme: "default"` so that it can easily be changed.

PR #425, opened 2018-11-09
Add delete and backspace to in-game help

pliuphys:
I added description of backspace and delete to the in-game control help. I only edited the display string files for English. So the other two has to be edited.


featurecat:
Thanks. This is totally unrelated to your PR but the Travis build is failing due to saving sgf comments being on by default. Maybe it will be better to disable that by default?

zsalch:
Yes, I think disable the 'append-winrate-to-comment' by default.

PR #436, opened 2018-11-12
Support replay the branch when press 'R'

zsalch:
by @phiming

Support replay branch when press key 'R', replay interval by option 'replay-branch-interval-seconds'


src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -1458,4 +1465,34 @@ public Color getBlunderNodeColor(BoardHistoryNode node) {
...
     }
   }
+
+  public void replayBranch() {
+    if (isReplayVariation) return;
+    int replaySteps = boardRenderer.getReplayBranch();
+    if (replaySteps <= 0) return; // Bad steps or no branch
+    int oriBranchLength = boardRenderer.getDisplayedBranchLength();
+    isReplayVariation = true;
+    if (Lizzie.leelaz.isPondering()) Lizzie.leelaz.togglePonder();
```
featurecat: Is it possible to have R, but NOT pause ponder? If not, lgtm

featurecat:
This change makes me happy :)

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
zsalch: I remove the line, It seems OK.

featurecat: But what happens when the variation changes while the user is replaying? Can you save the current variation, in case it changes?

zsalch: The replay will stop when the mouse is moving.

featurecat: No - the variation might change due to Leela zero thinking of a better one, right? I don't want it to change while the user is viewing it.

zsalch: I got it and will test it.

featurecat:
status update?

Issue #328, opened 2018-07-15
Leela zero is not getting loaded


someone235:
When I start lizzie I get the message "Leela zero is loading...", but it doesn't ever get loaded, and I have no error message.
I use Xubuntu 18.04

klaas1979:
I got the same problem using Mac OS X. I have leela-zero build it is working (copied it to lizzie directory as in the readme). I can force play via Enter-key. I edited the config file to get the outputs from leelaz and get stuff like:
```
Klaas-MBP:lizzie0.5 klaas$ java -jar Lizzie.jar 
> 1 version
> 2 lz-analyze 10
=1 0.15
> 3 time_settings 0 2 1
> 4 genmove B
=3 
=4 D4
> 5 time_settings 0 2 1
> 6 genmove W
=5 
=6 Q4
> 7 time_settings 0 2 1
> 8 genmove B
=7 
=8 D16
> 9 time_settings 0 2 1
> 10 genmove W
=9 
=10 Q17
```

But the message "Leela Zero is loading..." never disappears.

I cannot toggle "pondering" via space nor the options f, w, g and v.

featurecat:
Really strange... and the weight file is correct then? I should have a debug Config option. I can send you another version of Lizzie which will output more info. I can do this by tomorrow night I guess. I just got back from England tonight. Let me know if you solve it in the meantime.

klaas1979:
It is very strange, I did not figure it out, so cannot give concrete cause or directions :( I cannot recreate the problem and the solution. What I did:
1) Checkout latest lizzie and build lizzie via maven (mvn package)
2) Clean Checkout of leela-zero with the commit dae7011c1c8b03d4554410230015dde84118a9b7 (latest commit did not work leelaz is crashing)
3) using the provided network from your release package or the newest network with Hash 050375ce
==> it is working

So nothing to do for you but still strange. @someone235 maybe you can try it as stated above.

someone235:
I configured Leela Zero to work with CPU. Can it be related?

humuyan:
same problem +1
However pressing enter to force leela zero move works. No error message.

fallcat:
@humuyan @klaas1979 @someone235  Have you solved the problem? I can't get mine working on Mac and pressing enter to force leela zero works too.

someone235:
No, I didn't solve it

OlivierBlanvillain:
@someone235 Anything showing up in the console?

streetplay:
[...]

PR #442, opened 2018-11-14
Fix NPE from getLocationOnScreen_NoTreeLock

OlivierBlanvillain:
This commit fixes the exception reported by @seopsx in #328.

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
+++ b/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
@@ -1286,7 +1286,6 @@ public class LizzieFrame extends JFrame {
             new BufferedImage(scrollPane.getWidth(), scrollPane.getHeight(), TYPE_INT_ARGB);
         Graphics2D g2 = cachedCommentImage.createGraphics();
         scrollPane.doLayout();
-        scrollPane.addNotify();
         scrollPane.validate();
         scrollPane.printAll(g2);
         g2.dispose();


zsalch:
@OlivierBlanvillain 
If remove the `scrollPane.addNotify();`, the comment panel will draw wrong when mouse wheels:
image

OlivierBlanvillain:
`addNotify` is not supposed to be called in user code, we need to find another way to do this...

[javadoc](https://docs.oracle.com/javase/6/docs/api/java/awt/Component.html#addNotify%28%29):
addNotify
Makes this Component displayable by connecting it to a native screen resource. This method is called internally by the toolkit and should not be called directly by programs.

zsalch:
Yes, but it is a actual effective function.
Our final goal is to avoid current implementation of the comment.
Before that, I first tried to install Ubuntu to reproduce this error.

OlivierBlanvillain:
> Our final goal is to avoid current implementation of the comment.

I assume you mean "to avoid breaking the current implementation...".
Sure, I totally agree. I think we should really stop merging PRs without reviewing them...

featurecat:
Personally it seemed perfectly fine to me... I looked over it

OlivierBlanvillain:
Also seemed fine to me, this is why it's best to have the person who touched the code last to have a look :smile:

zsalch:
@OlivierBlanvillain 
> Our final goal is to avoid current implementation of the comment.
I actually mean that best way is change the method of the draw the comment of the sgf.
It takes some time, so I'm trying to solve the problem under Linux.

zsalch:
I run the lizzie 0.6 on the Ubuntu 18.04 and Jdk 1.8.0_191, it seems ok:
image

OlivierBlanvillain:
@zsalch It does run, but there is an exception printed in the console.

zsalch:
No exception in my console, maybe the jdk version is different.

Issue #443, opened 2018-11-16
the key to show/hide the winrate suggestion stones


PARKTAIJE:
Hi!
Thank you for the Lizzie project.

And I think it's better if there is the key to show/hide the winrate suggestion stones.
I think It could help more powerful thinking of playing games.
Now, pondering off is not hiding the winrate suggestions

Thanks.

kaorahi:
z or Z ? (cf. #317, #223, #157)

PARKTAIJE:
Yes! ,it is.
The key is what I wanted.
I didn't know the key z, Z exist..
I couldn't find it when holding x.
thank you, kaorash.

PARKTAIJE:
> z or Z ? (cf. #317, #223, #157)

featurecat:
good idea - we could add that inverted option to the Config. I think a lot of people will like it. also we must add z to the x menu (which is getting so crowded)

PARKTAIJE:
Thanks, kaorahi!
I have applied your commits about key Z,z.
It worked greatly.
Thank you very much, kaorahi!

PR #445, opened 2018-11-17
Toggle suggestions temporarily by z key

kaorahi:
The first commit implements #443 (toggle suggestions temporarily by z key).

The second commit adds z and Z into x menu.
Please feel free to correct/improve the descriptions.


src/main/java/featurecat/lizzie/gui/BoardRenderer.java
```diff
@@ -1205,6 +1205,14 @@ private boolean showCoordinates() {
...
   }
 
+  private boolean showBestMoves() {
+    return Lizzie.config.showBestMoves || Lizzie.config.showBestMovesTemporarily;
+  }
+
+  private boolean showBranch() {
+    return Lizzie.config.showBranch || Lizzie.config.showBestMovesTemporarily;
+  }
+
```
zsalch: Is it better to move these two methods to config.java?

kaorahi: They were moved to Config.java.

PR #447, opened 2018-11-19
Revert "Fix NPE from getLocationOnScreen_NoTreeLock"

OlivierBlanvillain:
Reverts featurecat/lizzie#442

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
+++ b/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
@@ -1275,6 +1275,7 @@ public class LizzieFrame extends JFrame {
             new BufferedImage(scrollPane.getWidth(), scrollPane.getHeight(), TYPE_INT_ARGB);
         Graphics2D g2 = cachedCommentImage.createGraphics();
         scrollPane.doLayout();
+        scrollPane.addNotify();
         scrollPane.validate();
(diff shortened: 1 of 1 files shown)


featurecat:
sounds good

featurecat:
what's the status. are you waiting for @zsalch? I would merge this if it's a problem.

zsalch:
> what's the status. are you waiting for @zsalch? I would merge this if it's a problem.

This seems to be only a issue in the openjdk, the official jdk is no problem.

zsalch:
If you want to solve this problem, I think it is necessary to separate the current panel. The board needs to be refreshed instantly, but other panels do not need it. Currently, the entire frame is refreshed, which causes many problems and limits many functions.
I'm already trying to make changes, and I'm currently finishing part of it. Maybe I will continue to push forward these two days and make a working version for everyone to discuss.

featurecat:
Ok. I'll wait on merging this, then. Maybe we need to divide the UI into separate panels, because I am hoping to add a toolbar on the UI anyway. Which would probably want to be separate from the canvas object.

featurecat:
Awaiting your solution.

zsalch:
This PR can be merged, because it will restore previous wrong fix.

About the new panel, I'll submit a new PR.
I use the JWindow as the panel, because it can be drag to other monitors.

Issue #321, opened 2018-07-11
Winrate heatmap


Gurxtav:
I'd like a feature where Lizzie lets LeelaZ explore each legal move on the board with a specified number of visits and then displays a heatmap with colors based only on calculated winrate for each move (the difference in winrate between this position and the one after the move or something like that).

This may be a good way for me to understand good direction of play (if there are a lot of green moves on the top of the left side for example).
Also, good moves and tesujis which the policy network missed can be found by exploring every move.

Of course it would be a bit computationally expensive since there are many legal moves, but something like 100 visits per move still gives good strength and can be computed in reasonable time.

featurecat:
oh that is pretty interesting. I imagine there would be a lot of noise with low playouts. Even 100. probably we can do this by executing the specified number of playouts on each move and writing the winrate

WYVYW:
MiniGo has something similar. The lower right corner shows the value of the board.
![image](https://user-images.githubusercontent.com/38362247/42717514-dc06438a-8701-11e8-9f82-eefb9cafd22d.png)

OlivierBlanvillain:
I think this would require some non trivial modifications on the Leela Zero side. Please reopen if/when implemented there.

Diapa:
no not at all.
It is probably possible to make it better by modifying LZ, but what I am talking about is a pure Lizzie thing.
Lizzie can just sutomatically go into every move on the board, let LZ think a little, note the winrate and go back, just like I can do this manually. 
And then display it as a heatmap when every move has been looked at.

alreadydone:
There's LZ review mode https://github.com/gcp/leela-zero/issues/1758#issuecomment-416186612 developed by @AncalagonX. He has added additional configurable options since then, but it seems he's not working on a pull request. You can use his release though.

gcp:
>we can do this by executing the specified number of playouts on each move and writing the winrate

I agree with this. Note that you can now control the number of playouts/visits on the fly via lz-setoption.

>Changing that to "automatically go into every move on the board" would require large architectural changes.

That's unfortunate :-(

Nevertheless it makes no sense from my side to implement something protocol/engine side when it can already be achieved by the GUI. Those kind of protocol design mistakes have the unfortunate habit of potentially outliving both Leela Zero and Lizzie, which is probably why we're in the state right now where people regularly ask to make the time control parameters settable via command-line parameters...

Let's hope we can encourage people to improve Lizzie instead.

gcp:
[...]

PR #451, opened 2018-11-21
Option to color moves by winrate instead of visits.

TFiFiE:
As it might soon become possible to manipulate the search of Leela Zero on the fly (gcp/leela-zero#1949), generating a winrate heatmap (#321) could become feasible to implement on the front-end side.

In anticipation of this, this change makes it possible for moves to be colored by winrate instead of visits, either by pressing `E` or through a config setting. For consistency sake, the color mapping in both modes is approximately "perceptually" linear with respect to the respective statistic (with maximized saturation). If the displayed winrate is fixed to be from black's point of view, then winrate coloring mode will follow this.
As the most visited move can now possibly appear in red, not having the highest winrate will result in it getting a magenta ring instead (in both modes, again to keep things consistent).

Translations still need to be added.

This is how it can look after modifying LZ to do a wide search:
screenshot


TFiFiE:
Instead of getting a linear mapping between winrate and color, it seems more useful to be able to distinguish winrates closer to equality better than further away:
screenshot

AncalagonX:
This is definitely a toggle that should be added to Lizzie. Normally, when using unmodified LZ in Lizzie, visit-based coloring is the most appropriate mode. But since I am using a custom LZ that lets me modify search width, coloring by visit count is meaningless. I did not test this PR, but I implemented the same thing on my personal Lizzie builds, and made many similar discoveries/decisions that @TFiFiE did regarding coloring.

I ended up using sharper values for hue and alpha scaling factor. I recommend a sharper alpha scaling so that all the bad red moves aren't so bright and _overwhelming._

    final double HUE_SCALING_FACTOR = 0.9;
    final double ALPHA_SCALING_FACTOR = 3.0;

The red-to-green gradient scales max green based on the highest winrate Lizzie sees on the board each move. An example from my implementation with these scaling values:
2018-12-13_23-28-19_ javaw

Another example where black must respond to a forcing move, and how the winrate-based colors look when scaled in this manner:

2018-12-13_23-58-22_ javaw

Here's [my five-line commit](https://github.com/AncalagonX/lizzie/commit/50d85505d8dbf1c69f9553c31e4f586b0d7f06bc) where I implemented winrate-based coloring in case it's helpful, though it's based on an older version of Lizzie.

zsalch:
This idea is very good.
But when by winrate displaying, the color seems not as clear as previous, I still tend to reserve the previous color theme when by winrate.

by winrate vs by visits:
image

TFiFiE:
By coloring moves based on absolute instead of relative evaluation, it is made clear on a glance what the resulting winning chances would be for each of the evaluated moves, which is what the game is ultimately about. On the left, it's immediately obvious that, in the first position, each of the evaluated moves would keep the game close to even, whereas the second position is clearly losing regardless of what would be played. The near-identical coloring therefore makes sense in both cases.

featurecat:
I will try to get this into 0.7.

featurecat:
@TFiFiE that's a good point but for some reason I don't like it. It might have to do with emphasizing the difference between each of those moves. I definitely want to see se green on the board and some red, to differentiate. If every move looks similar it looks strange. I don't know, it could be an option users can enable.

featurecat:
I added what I feel is a more Lizzie-like color theme. Thoughts?
image
image
I'm not sure whether or not I should highlight the best move in cyan.

featurecat:
For now I'm merging this PR + my changes into master. It's currently bound to 'K' without a display in X. I am hoping to make changes to how the X show controls menu works, so I will probably put this into a toolbar/menu.

zsalch:
I feel that should still retain cyan as the best move. After all, this has become a feature of Lizzie.

gogre:
another opinion on move colouring, plus a few other things:
https://github.com/gogre/irgo/blob/master/musicalstones.md

Dorus:
I'm not sure if this is the best spot to share my random ideas, but hopefully it provide some insight or invokes some discussion.

Regarding coloring: I would prefer the color scheme to more represent how much of a blunder a certain move it, making everything above 50% green and below 50% red, and having an almost undistinguishable cyan and greed between two moves that are 54% and 51% (a blunder over 3%) seems bad. I would prefer some clear highlight for moves within 1-2% of the best move, and then some sliding color into red for the rest.

Regarding search: Be careful on how you manipulate search. I've seen screenshots where all moves got 100 visits, and that will work. However if you let search free after that, you can still get problems. During leela search, you can have moves that start at 80% win (for example), but then fall down to 20% win after a deeper search. However what happens in practice is that the winrate drops off to 75% before another 80% winrate move takes priority, keeping the 75% winrate move and 75% with relative high playouts, 
When tweaking the search, i do not think it's efficient to give equal visits to all possible moves, but you might be able to use things like LCB to decide on moves that might require more visits for accurate winrate output, even if those arn't the best moves. As for analyses, you would not only want to know a move is not recommended, but also how bad it is.

PR #473, opened 2019-01-02
Updated readme with more getting started info

toomasr:
Based on my experience today setting Lizzie up I've updated the README with information to new comers.


featurecat:
Thanks. Sorry it took so long to merge. I will merge this now, but since I am probably changing the error handling and controls/UI for Lizzie 0.7, some of the instructions will need to change. I will work on those changes when we release!

PR #491, opened 2019-02-20
Fix #490 (Japanese font problem)

### LINKED ISSUE #490: The lastMove label is garbled in l10n
Hi,
The lastMove label is garbled in Japanese localization.
I have no problem with other labels.
Do you have any ideas?

wr

kaorahi:
OpenSans font was used for Japanese text.


featurecat:
Ah... thanks. Will the Config's default font support Japanese or will they have to manually change it?

kaorahi:
The default font seems to support Japanese in my environment.
I have no explicit font setting in config.txt.

    $ grep -ri font config.txt persist theme
    config.txt:    "comment-font-size": 0,
    theme/yasnaya/theme.txt:  "comment-font-color": [255, 255, 255, 200],

PR #501, opened 2019-03-18
next: Add the gtp console dialog

zsalch:
Add the Gtp Console Dialog (Press E to toggle and Auto Start if turn on the option 'print-comms'):

The style option:
"gtp-console-style": "body {background:#000000; color:#d0d0d0; font-family:Consolas, Menlo, Monaco, 'Ubuntu Mono', monospace; margin:4px;} .command {color:#ffffff;font-weight:bold;} .winrate {color:#ffffff;font-weight:bold;} .coord {color:#ffffff;font-weight:bold;}"

image


TFiFiE:
I don't see how this feature has a better claim over the 'E' key press than #451.

featurecat:
I will try to overhaul the UI keys so that you don't need to fight over the E key, haha. Don't worry. I will try to fix.

zsalch:
It doesn't matter which key is used here, after all, this feature is not used frequently. More situations are always on or off.

PR #502, opened 2019-03-21
next: Add the Avoid/Allow Dialog

zsalch:
Add the Avoid/Allow Dialog (Press Alt+A) for https://github.com/gcp/leela-zero/pull/1949

image


featurecat:
Oh I'm glad you did this already, I was planning to. I don't like the UI of this because I'd like users to be able to just click+drag on the board where they want to avoid/allow. I will reuse your code in order to make the changes but I won't merge the UI, sorry. Leaving this PR open until I implement the feature.

zsalch:
Yes, this UI is just usable and requires a better way of interacting.

featurecat:
I played around with it, it is quite fun. I think it will be really nice to drag the mouse over areas as a way of interacting with it!

peko19:
> I played around with it, it is quite fun. I think it will be really nice to drag the mouse over areas as a way of interacting with it!

https://github.com/leela-zero/leela-zero/issues/2380#issuecomment-491131296

PR #503, opened 2019-03-21
next: Save/Restore  the Window and Board Position in the persist

zsalch:
Save/Restore the Window and Board Position in the persist

### DIFF
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -73,6 +73,7 @@ public class Config {
   public Optional<Map<Double, Color>> blunderNodeColors;
   public int nodeColorMode = 0;
   public boolean appendWinrateToComment = false;
+  public int boardPositionProportion = 4;
   public String gtpConsoleStyle = "";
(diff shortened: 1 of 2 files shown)


src/main/java/featurecat/lizzie/Config.java
```diff
@@ -379,7 +381,7 @@ private JSONObject createDefaultConfig() {
     ui.put("autosave-interval-seconds", -1);
     ui.put("handicap-instead-of-winrate", false);
     ui.put("board-size", 19);
-    ui.put("window-size", new JSONArray("[1024, 768]"));
+    ui.put("window-size", new JSONArray("[1229, 768]"));
```
featurecat: ?

zsalch: The design of the Lizzie's UI is actually designed in a 16:10 ratio. The previous default [1024,768] is 4:3, which seems to be narrow on both sides.

featurecat:
I'll think about it because actually, default window size is too big for my screen. should be screen dependent.

PR #504, opened 2019-03-21
next: Show the visits/second in the title bar

zsalch:
Show the visits/second in the title bar:

image


featurecat:
Later we may want to optimize the display of this number for large values e.g. 5.2k instead of 5235.

PR #513, opened 2019-03-31
next: New UI - Multiple independent panels that can be dragged out

zsalch:
I tried to develop a new UI layout that used multiple independent panes that can be dragged out.

Split the original single frame into the following panes:
Main Board
Basic Info
Winrate
Sub Board
Variation Tree
Comment
image

This will refresh the individual Pane.
I wrote a simple layout manager that currently only implements the current layout and needs further optimization.

At the same time, each Pane, like the JToolBar, can be dragged out to be a separate window so that we can display them on different displays. The window will restore a pane when the window be closed.
image

Possible problems and next steps:

1. Need a better background implementation;
2. Need a better layout manager;
3. It is best to optimize the current source structure;
4. Need more tests to find hidden problems.

This is a test version that can be run and looks forward to a better implementation.


featurecat:
Wow, cool! Even if we don't finish the fully correct implementation, it can be a Config setting. I'll take a look

featurecat:
I am getting the compile error: Error:(41, 17) java: package sun.swing does not exist, in BasicLizziePaneUI

Do you know what is wrong?

zsalch:
I just temporarily removed the package sun.swing, please try again.

featurecat:
It's really cool, but, appears quite glitchy right now. In the backgrounds of each component now, I can see a glitched version of the board.. and the Lizzie backgrounds etc aren't working correctly. Key commands aren't working (for example I can't use the up arrow to undo). Dragging anywhere moves the pane, when that's not desirable. Note that in 0.7 release, dragging on the board will be the way you enable avoid-moves and search-moves from your other PR.

Is it possible to fix some of these issues? I want to include it as an optional but nondefault feature. 

image

zsalch:
1. It is not dragable by default. It can only be dragged when you press a special key (for example: Alt+W). The panels will change to the design mode. After the drag is complete, press (Alt+W) again to exit the design mode.

2. Fixed the Key Issue.

3. I will check the background issue.

featurecat:
it was draggable by default for me...

zsalch:
I adjusted the background display issue. It seems to have improved.

zsalch:
> it was draggable by default for me...
This is a bit strange, I have commented the event registration code.

zsalch:
Fixed the Background Issue.

zsalch:
Save/Restore the Window & Pane Position

zsalch:
Add Background Process.

zsalch:
Can turn off design mode when press Alt+W in the LizziePane.

zsalch:
Add the status indicator:
image
image

featurecat:
Interesting idea but the icon has to be self explanatory. I'm not sure what the best way is either. I'll test this PR tonight, thanks for all the updates. Hopefully everything is fixed.

gogre:
"icon has to be self explanatory" - indeed, yes. However, if the information being communicated intrinsically requires more pixels than an icon provides, that is impossible.  But an icon can be memorable, even if not intrinsically mnemonic.  Well-known examples are the Star of David, the Christian crucifix, the crescent and star of Islam, and the hammer and sickle of Russia..  None is self-explanatory, but each is so sinple and distinctive that anyone can remember them to know what they mean when they see them.  Another example is the Roman alphabet - although its iconography is less well thought out than religious symbols, because there are several cases where two different icons have similar shapes - eg 5 and S, 0 and O.

In these cases, the icon signifies a constant, but the icon for win% must signify the value of a variable on a linear scale of 0 to 100.  The most obvious and mnemonic way to illustrate this is by analogy with a ubiquious concept: vitality.  And that, in turn, is most meaningfully expressed by analogy with skin colour, or leaf colour.  Or the colour of the sun.

When someone is hot, they flush red from increased blood pressure - and when cold, their skin turns blue as blood is drained from the extremities.

A vibrantly alive leaf is bright green, but as it dies, it turns first yellow, then orange, then brown.

The sun, at its zenith, is bright yellow - and at its nadir, near the horizon, it fades into red.

So whichever colour scale you choose, it will implicitly have one or more of these interpretations, whether or not you intend it.

The BBC once flirted with an icon called a "swingometer" to dramatise the swing from one party to the other as election results came in.  It was cute, but suffered from the drawback that a real swingometer, a pendulum, naturally gravitates towards the middle.

The thermometer, which uses an expanding or contracting coumn of mercury to indicate temperature was once ubiquitous, but since the advent of the digital age, it is starting to lose its mnemonicity.

In Go, there are two players, two colours: black and white.

So a solid black icon would imply that black dominates, a white one that white does, and a grey one that there is no clear leader.

So that's all you need.  You could have 50 shades of grey if you wanted to be pedantic.

featurecat:
The panels look strange due to no background transparency. Is there a way to make them at least have the same background texture? Or to look normal when they are not in the window pane mode?

image

featurecat:
@gogre Please make your suggestions in a separate issue. And, please make them more concise and readable.

zsalch:
I will try to append background transparency if this is necessary.
Also, are there any screenshots for me to understand?

featurecat:
I think if the background textures are not visible, it will make some nice themes look uglier. It would be best to make them transparent.

featurecat:
Yes the screenshot is just to show that my backgrounds were not rendering the same as before.

zsalch:
Added the pane background transparency process.

zsalch:
Update Pane Background Transparency Process & Fixed Draw Controls.

PR #515, opened 2019-04-08
Apply changes made by fmt-maven-plugin

apetresc:
I'm preparing a couple of PRs and it's very annoying that the `com.coveo.fmt-maven-plugin` that was added to Lizzie in late October keeps modifying these 9 files every time the project is built, which messes up all the diffs.

Presumably if that format plugin is installed, it's expected for the checked-in source to conform to it. So here are those 9 changes applied, so all future diffs will be nice and clean :)

### DIFF
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -6,10 +6,8 @@ import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.*;
-
-import org.json.*;
-
 import javax.swing.*;
+import org.json.*;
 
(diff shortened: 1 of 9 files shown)


featurecat:
What changes are you working on?

apetresc:
@featurecat Firstly #517, and preparing something similar for Linux :) My goal is to prepare Lizzie for packaging for many common OSes and distributions from however packages are natively distributed on those platforms (things like Homebrew, AUR, Ubuntu PPAs etc), rather than expecting users to have to run JARs manually.

(I am doing similar packaging for leela-zero itself)

featurecat:
Sounds great

PR #517, opened 2019-04-08
Automated creation of Mac .app bundle

apetresc:
For Mac users, it is a bit of a pain to run `.jar` files directly. You get an ugly Java logo in your dock, you can't associate SGF files with them, you can't open them with app launchers, etc. This PR adds a Maven build target, `appbundle:bundle`, which generates a standalone `Lizzie.app` that you can distribute anywhere. It uses the new Lizzie logo and runs like this:

(Notice the dock icon entry on the left and the app in `/Applications` on the right).

This `Lizzie.app` would also be appropriate for distribution as part of a new Lizzie release on GitHub, and would allow it to be easily packaged in Homebrew-Cask so users can just `brew cask install lizzie` from anywhere.


featurecat:
You get a Java logo even in the current master branch? I (wrongly?) assumed it shows the logo in Mac.

featurecat:
Will Mac users still need to compile Leela Zero, or is it magically handled?

apetresc:
Yup, it's still a Java icon in `master` at the moment.

apetresc:
Yup, the `.app` bundle is a compiled package; Mac users just need to double-click it, no compilation or terminal involved at all.

featurecat:
Ew, ugly, ok. I don't have a Mac so I really appreciate you making this.

featurecat:
How would Mac users switch out new networks, or new Leelaz executables?

apetresc:
> How would Mac users switch out new networks, or new Leelaz executables?

Good question.

So, the working directory of the .app is set to the root directory of the .app itself, which technically a user _can_ access by right-clicking the .app and selecting "Show Package Contents", like this:

(At which point they can modify `config.txt` and paste in new `network.gz` etc)

However, this is largely not necessary anymore. Since Leela-Zero 0.16, `leelaz` will automatically use the network in `$XDG_CONFIG_HOME/leela-zero/best-network` without any `-w` parameter at all. For most users I expect this is how they configure `leelaz` in every tool. I think it would be a good idea for Lizzie (on _all_ platforms) to omit the `-w` parameter from the default-generated `config.txt` if that directory is in use, since it is the standard solution for Leela-Zero in general.

On the Mac specifically, though, I am currently pushing to get Leela-Zero included in the core Homebrew repository (progress [here](https://github.com/Homebrew/homebrew-core/pull/37366) and [here](https://github.com/Homebrew/homebrew-core/pull/38743)). That formula installs a network to `/usr/local/opt/leela-zero/share/leela-zero/` (which is where Homebrew expects all such resource files to be), so we can use that as a fallback default location. I can add that logic to this PR!

Once Lizzie pushes a tagged release that includes this App bundle, we can similarly get Lizzie included in Homebrew; after that, Mac users will be able to literally just do:

```$ brew install leela-zero lizzie```

and in a few seconds have both things running together with a recent network, without any configuration at all.

(This is currently already the case in Arch Linux through the [https://aur.archlinux.org/packages/leela-zero](`leela-zero`) and [https://aur.archlinux.org/packages/lizzie](`lizzie`) packages in AUR!)

zsalch:
I will also complete the config dialog, which will allow more users to simply modify the settings.

featurecat:
I will need some help writing Lizzie installation+usage+customization then, for mac.

PR #518, opened 2019-04-09
support 0.17 engine 

yzyray:
if use 0.17 engine it can now play aginst leelazero and when analysing it shows lcb instead of winrate.
thanks for featurecat's advice I adjusted the code it Can now support both 0.17 engine(use lcb tag)and 0.16 0.15 engine(use winrate tag)


featurecat:
Thanks for addressing my concerns earlier.

src/main/java/featurecat/lizzie/analysis/MoveData.java
```diff
@@ -29,6 +35,7 @@ public static MoveData fromInfo(String line) throws ArrayIndexOutOfBoundsExcepti
     // Todo: Proper tag parsing in case gtp protocol is extended(?)/changed
     for (int i = 0; i < data.length; i++) {
       String key = data[i];
+      boolean islcb = false;
```
zsalch: Should be moved to the front of the for.

src/main/java/featurecat/lizzie/analysis/MoveData.java
```diff
@@ -73,6 +104,9 @@ public static MoveData fromSummary(String summary) {
 
   private static Pattern summaryPattern =
       Pattern.compile("^ *(\\w\\d*) -> *(\\d+) \\(V: ([^%)]+)%\\) \\([^\\)]+\\) PV: (.+).*$");
+  private static Pattern summaryPatternold =
+      Pattern.compile("^ *(\\w\\d*) -> *(\\d+) \\(V: ([^%)]+)%\\) \\([^\\)]+\\) PV: (.+).*$");
+  // support 0.16 0.15
```
zsalch: Should be check both pattern.

featurecat:
Ok I've tested it out. There's a problem: LCB value starts out at 0.0, and takes several playouts to converge close to the winrate. On cpu-only mode, this takes a while. But it will be slightly below the winrate, so the winrate graph bounces up and down due to LCB being lower than winrate.

Do you have a solution in mind? I will try to think of one, too.

By default, Lizzie handles LCB well, except when the visit count is low. Then you get a cyan move that is transparent. A simple fix would be to make cyan moves have an opacity of 1.0. Compared to that, is there an advantage to using LCB instead of winrate?

yzyray:
LCB is more accurate than winrate, espacially when some secondary point's playouts is very small but it can still have high winrate it won't appear in lcb.  for example  we  see two points ,one playouts 2000 and winrate 55.5%, another one playouts 1000 and winrate 55.8%, in  this time who's lcb is higher usually will be better . when playouts is bouts 1000 choose a move by maxlcb vs choose a move by  maxplayouts maxlcb can win more than 60% games.

featurecat:
OK, then I'm willing to merge this. But I want LCB value stored in addition to the winrate value. Because it will make ctrl+x only switch the value displayed, not the value stored.

Also I want winrate to still be the default. That's because, Lizzie already displayed variations and cyan best moves with LCB.

And below 20 playouts on a move let's display the winrate instead of the lcb value. By then it should be roughly stabilized so we can use LCB.

src/main/java/featurecat/lizzie/gui/ConfigDialog.java
```diff
@@ -530,6 +549,8 @@ protected DocumentFilter getDocumentFilter() {
                   });
         });
     txtMaxAnalyzeTime.setText(String.valueOf(leelazConfig.getInt("max-analyze-time-minutes")));
+    txtAnalyzeUpdateInterval.setText(
+        String.valueOf(leelazConfig.getInt("analyze-update-interval-centisec")));
     txtAnalyzeUpdateInterval.setText(
         String.valueOf(leelazConfig.getInt("analyze-update-interval-centisec")));
```
zsalch: It seems duplicate.

src/main/java/featurecat/lizzie/gui/ConfigDialog.java
```diff
@@ -679,8 +702,33 @@ public boolean isWindows() {
...
+    if (rdoWinrate.isSelected()) {
+      return false;
+    }
+    return true;
+  }
+
   private void setBoardSize() {
-    int size = Lizzie.config.uiConfig.optInt("board-size", 19);
+    int size = Lizzie.config.uiConfig.optInt("board-size", 0);
     txtBoardSize.setEnabled(false);
```
zsalch: Why change 19 to 0?

yzyray: ...maybe by mistake , and now i find my local code is 19...

src/main/java/featurecat/lizzie/gui/ConfigDialog.java
yzyray: It seems so...

yzyray:
LCB value stored in addition to the winrate value is great. yesterday i replayed a game with it. i find  a transparent cyan move maybe not a bad thing,it shows the move is best but dont have enough playoyts,so probably there is as least another good move.

src/main/java/featurecat/lizzie/Config.java
```diff
@@ -355,6 +355,8 @@ private JSONObject createDefaultConfig() {
     leelaz.put("max-game-thinking-time-seconds", 2);
     leelaz.put("print-comms", false);
     leelaz.put("analyze-update-interval-centisec", 10);
+    leelaz.put("show-lcb-winrate", true);
+    leelaz.put("leela-version", 17);
```
zsalch: It is best to automatically determine whether to support lcb

src/main/java/featurecat/lizzie/Config.java
```diff
@@ -355,6 +355,8 @@ private JSONObject createDefaultConfig() {
     leelaz.put("max-game-thinking-time-seconds", 2);
     leelaz.put("print-comms", false);
     leelaz.put("analyze-update-interval-centisec", 10);
+    leelaz.put("show-lcb-winrate", true);
```
zsalch: 'show-lcb-winrate' reserved

src/main/java/featurecat/lizzie/analysis/Leelaz.java
```diff
@@ -315,6 +316,12 @@ private void parseLine(String line) {
           String[] ver = params[1].split("\\.");
           int minor = Integer.parseInt(ver[1]);
           // Gtp support added in version 15
+          try {
+            Lizzie.config.leelazConfig.putOpt("leela-version", minor);
+            Lizzie.config.save();
+          } catch (IOException e) {
+            e.printStackTrace();
+          }
           if (minor < 15) {
```
zsalch: Remove the change.

src/main/java/featurecat/lizzie/analysis/MoveData.java
```diff
@@ -18,14 +19,22 @@ private MoveData() {}
...
    * @param line line of ponder output
    */
   public static MoveData fromInfo(String line) throws ArrayIndexOutOfBoundsException {
     MoveData result = new MoveData();
     String[] data = line.trim().split(" ");
-
+    boolean islcb =
+        (Lizzie.config.config.getJSONObject("leelaz").getInt("leela-version") >= 17)
+            && Lizzie.config.config.getJSONObject("leelaz").getBoolean("show-lcb-winrate");
     // Todo: Proper tag parsing in case gtp protocol is extended(?)/changed
```
zsalch: It is best to automatically determine whether to support lcb.

src/main/java/featurecat/lizzie/gui/ConfigDialog.java
zsalch: Please remove.

src/main/java/featurecat/lizzie/gui/ConfigDialog.java
zsalch: Please fix.

PR #536, opened 2019-05-18
Using Zen to estimate scores

yzyray:


featurecat:
Sorry but we do NOT have permission to use Zen. It's not open source, it's
a program that costs a lot. They would definitely be angry if we used Zen
without permission.

We should contact them first. If they approve (probably not), then this is
great. If not we can always try other software. I don't want to introduce
pirated software into Lizzie.

yzyray:
OK, then i close this pr and try to use gungo if it is possible

featurecat:
There's a chance they might say yes? Keep it open for now at least and make
a separate one for other ai.

I don't think gnugo is strong but at least it's an estimate.

tychota:
I think my-lizzie use gnugo and zen for estimate. 

If you have bought zen, it use zen (require you to copy some dll). Else it fallback to gnugo.

featurecat:
That would be a nice alternative. I don't want to redistribute the dll.

tychota:
Completly agree with not redistibuting dll but the rest of the code can be worthy :)

yzyray:
I delete zen.dll but keep YAZenGtp.exe  it has many versions,if not use the right version will not working

zsalch:
It is recommended to remove all exe and dll files, leaving only the interface-related code files. If the user has a related program, can use this interface.

src/main/java/featurecat/lizzie/gui/LizzieFrame.java
```diff
@@ -1569,4 +1577,34 @@ public void run() {
...
+    isCounting = true;
+  }
+
+  public void noAutoCounting() {
+    this.isAutoCounting = false;
+    if (Lizzie.config.showSubBoard) Lizzie.frame.subBoardRenderer.removeCountBlock();
+    else Lizzie.frame.boardRenderer.removeCountBlock();
+    Lizzie.frame.repaint();
+    this.countResults.isAutocounting = false;
+    this.countResults.button2.setText("自动判断");
```
zsalch: "自动判断" -> DisplayStrings.properties

yzyray: forget this one .....fixed

yzyray:
Removed the YAZenGtp.exe and if anyone want to test it.may find it in my branch "menu"

Issue #541, opened 2019-06-20
Is the next version coming with native KataGo support


hydrogenpi:
Is the next version coming with native KataGo support

zsalch:
Yes.

PR #542, opened 2019-06-21
Support KataGo (#541) and its score estimation

kaorahi:
* Accept KataGo without the version cheat "-override-version 0.17"
* Show the estimated score above the winrate bar


src/main/java/featurecat/lizzie/Config.java
```diff
@@ -34,6 +34,7 @@
   public boolean showCoordinates = false;
   public boolean colorByWinrateInsteadOfVisits = false;
   public boolean showLcbWinrate = false;
+  public boolean isKataGo = false;
```
zsalch: Great job! Is it possible to detect dynamically?

kaorahi: It is detected dynamically from the GTP response to "name" command just before the version check. Do we need more?

featurecat:
Thanks. Can you let it use KataGo without a cmdline argument? I want it to
be very easy to setup.

kaorahi:
You only need to modify two lines in config.txt to use KataGo. No cmdline argument is required. Lizzie detects KataGo automatically through GTP.

~~~
"network-file": "katago-network.gz",
"engine-command": "./katago gtp -model %network-file -config gtp.cfg",
~~~

Yakago:
I tried to compile with this change, and used @alreadydone 's windows compilation of KataGo (https://github.com/lightvector/KataGo/issues/2)

I had an issue that lizzie would still use lz-analyze. I suspect that it is because it takes quite some time for the engine to load, and therefore respond with the proper 'name'. I'm not too sure as I'm a bit inexperienced will all this.

I just made some manual changes to force it to use kata-analyze and draw the score difference no matter what..

src/main/java/featurecat/lizzie/Config.java
zsalch: I think dynamic detection is necessary because the user might configure multiple engines and switch between them.

kaorahi:
@Yakago, how about b320284?

Anyway, we may have another better implementation soon :)
https://github.com/lightvector/KataGo/issues/25#issuecomment-505276460

Yakago:
I just downloaded https://github.com/kaorahi/lizzie/tree/support_katago :)

Yakago:
I tried again, and now I get it to start with kata-analyze

src/main/java/featurecat/lizzie/Config.java
zsalch: I think it would be best to move config.isKataGo to Leelaz.isKataGo for multiple engine.

zsalch:
LGTM. Maybe it will be merged together #546 .

PR #543, opened 2019-06-24
next: Support KataGo non-square and big boards

zsalch:
Support KataGo non-square and big boards.

image
image
image
image


zsalch:
For subsequent PR, merge this PR first.

l1t1:
what are the min and max board size?

PR #544, opened 2019-06-24
next: Support the engine preload

zsalch:
Press Ctrl+X to check the pre-load checkbox to preload the engine.
image


zsalch:
Since many subsequent modifications are based on this PR, it is merged first.

PR #550, opened 2019-07-02
Draw suggestion text after next-move ring (fixes #549)

### LINKED ISSUE #549: "Next move" ring drawn on top of move stats
The rings indicating the next move(s) in the variation tree is drawn after suggestions are drawn. With KataGo, there are potentially three pieces of information in the stats (winrate/visits/estimate) and it is easy for them to slightly exceed the bounds of the circle, but then they are partially obscured by the ring being drawn on top of them.

For the nicest aesthetics, the colored disc should be drawn first, then the next-move ring, then finally the text stats on top.

dfannius:


zsalch:
Could you provide a screenshot for demo?

dfannius:
Here's a picture after the change. Previously, the white ring would be drawn over the text, making the `45.0` in particular hard to read.

Of course, there is still an issue with the black rings regardless of draw order.

Screenshot

src/main/java/featurecat/lizzie/gui/BoardRenderer.java
```diff
@@ -125,20 +134,26 @@ public void draw(Graphics2D g) {
...
       return;
     }
 
     if (!isShowingRawBoard()) {
       drawMoveNumbers(g);
       //        timer.lap("movenumbers");
+      List<TextData> textDatas = new ArrayList<>();
       if (!Lizzie.frame.isPlayingAgainstLeelaz && Lizzie.config.showBestMovesNow())
-        drawLeelazSuggestions(g);
+        drawLeelazSuggestionsBackground(g, textDatas);
```
zsalch: Why do need a drawLeelazSuggestionsBackground instead of drawing directly in drawLeelazSuggestionsForeground? Is there any problem?

dfannius:
I was notified by email of a comment but I can't find it here. In case it wasn't deleted intentionally: `drawNextMoves` and `drawLeelazSuggestions` are independent (each feature can be enabled separately), making it slightly awkward to combine them into a single function, and also `drawLeelazSuggestions` iterates over the whole board while `drawNextMoves` iterates over `getNexts()`.

In fact the current method of iteration in `drawLeelazSuggestions` appears to be extremely inefficient in principle, in that it iterates over every point of the board, and at each point looks through all of `bestMoves` to see if any of them happen to be at the current point. (There is a comment that it "looks better with shadows" so there's some reason for it.) I don't know if this actually does affect performance but it makes me cringe a bit. If we really need to iterate over the board in order, I would suggest sorting `bestMoves` by coordinate to start with, then iterating over the sorted list. In that case we could also sort the result of `getNexts()` and perform that iteration in parallel. But that's a separate refactor.

zsalch:
Maybe I didn't describe it clearly.
I mean, the function drawLeelazSuggestions is splited into two parts drawLeelazSuggestionsBackground and drawLeelazSuggestionsForeground.
Instead of simply moving the drawLeelazSuggestions to the back of drawNextMoves.
Is there any reason for this?

dfannius:
Do you mean do all of `drawLeelazSuggestions` after `drawNextMoves`? Then most of the ring drawn by `drawNextMoves` gets obscured by the colored disc drawn by `drawLeelazSuggestions`.

zsalch:
LGTM.

PR #554, opened 2019-07-05
Fixed and support more for the KataGo

zsalch:
1. Fixed some issues
2. Support change the komi
3. Support the KataGo handicap.


roy7:
This only works for the engine placing handicap stones. If the human is playing black with a handicap, then the engine is playing the stones for the human. Will try to make some change to issue a fresh PR if I can figure it out. Only issueing handica() when engine is black is easy, the trickier part is only changing blackToPlay to false after all handicap stones are placed if playerIsBlack.

PR #566, opened 2019-07-12
next: Change the current move indicator option

zsalch:
Change the current move indicator option

'solid-stone-indicator' -> 'stone-indicator-type' for more type:

0: No indicator
1: Circle
2: Solid


ivysrono:
@zsalch Where is this config? I build the master but do not find this config.

PR #567, opened 2019-07-12
Removing the duplicated 95% factor for scoreMean info

rexl2018:
Removing the duplicated 95% factor which made the scoreMean graph and the current scoreMean text in wrong position.


yzyray:
I added 95%, so the scoreMean line can always be a little short than winrate line, then the two lines  wont overlapping.

rexl2018:
Your code made the scoremean graph narrower in X coordination, which results in that the move 100’s scoremean graph is under current move white vertical line when your current move is move 95.

gjm11:
I find the mismatch between these two lines very strange. I appreciate the aim of making things more visibly separate from one another, but I think the cost is greater than the benefit. Maybe they should be two separate graphs, one above the other, though of course that reduces the amount of space available for each.

PR #572, opened 2019-07-13
Rightclickmenu for avoid/allow

yzyray:


src/main/java/featurecat/lizzie/Config.java
```diff
@@ -425,6 +425,7 @@ private JSONObject createDefaultConfig() {
     leelaz.put("print-comms", false);
     leelaz.put("analyze-update-interval-centisec", 10);
     leelaz.put("show-lcb-winrate", false);
+    leelaz.put("avoid-keep-vairations", 30);
```
zsalch: typo?

yzyray: fixed

PR #577, opened 2019-07-18
Add a simple menu

yzyray:


zsalch:
LGTM.

TFiFiE:
This change somehow breaks the communication between Lizzie and the engine.

yzyray:
> This change somehow breaks the communication between Lizzie and the engine.

When does it happen?

PR #591, opened 2019-08-03
Add showPolicy at toolbar

yzyray:
policy


lp200:
Can you add shortcut key?

yzyray:
> Can you add shortcut key?

added in pr (#596)

PR #600, opened 2019-08-07
Update board, white stone and Russian translation

ParmuzinAlexander:
https://imgur.com/ucE0HKB


ParmuzinAlexander:
Сommit with Russian translation added automatically!
This is beta. I'm not sure about the encoding, we need to check.

PR #624, opened 2019-10-02
Improve the way of display suggestions moves

yzyray:
I think it is not necessary to scan every coordinates on board,and there's some formatte differences,i dont know how to make it same.

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/BoardRenderer.java
+++ b/src/main/java/featurecat/lizzie/gui/BoardRenderer.java
@@ -106,6 +106,11 @@ public class BoardRenderer {
     int suggestionX;
     int suggestionY;
     boolean flipWinrate;
+    Color circleColor;
+    boolean hasMaxWinrate;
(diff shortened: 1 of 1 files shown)


featurecat:
This is a good idea and will help improve rendering speed, but it has a small issue. Now, the shadows that are casted by move suggestions are cast on top of existing suggestions. You should probably do two passes: one to draw all the shadows, and one to draw all of the suggestions. It will make them look better.

yzyray:
> This is a good idea and will help improve rendering speed, but it has a small issue. Now, the shadows that are casted by move suggestions are cast on top of existing suggestions. You should probably do two passes: one to draw all the shadows, and one to draw all of the suggestions. It will make them look better.
> 
> You can see this if you have two suggestions that are next to each other. One will cast a shadow on top of the other.

Yes,there is a shadow on top of the other, and i separate drawLeelazSuggestionsBackground into twp parts shadow and circle.Now it seems the same effects as before.

PR #641, opened 2019-10-15
Fix cannot delect Sgf file with capital extendsion ("*.SGF")

cngoodboy:
Like title, just an small fix.


timuzhti:
According to the [documentation for this class](https://docs.oracle.com/javase/10/docs/api/javax/swing/filechooser/FileNameExtensionFilter.html), the way things were before the change was the correct way, and that's how things seem to work on my system (OpenJDK 10, OpenSUSE Tumbleweed). Does the change actually fix things, and if so is there any way of fixing the issue without breaking it in other situations?

PR #650, opened 2019-10-31
Improve subboard and add option not refresh variation when displayed

yzyray:
Sometime the variation changes too fast on subbord and mainboard(when mouse is point at a suggestion move).So I make subbord stop refresh branch when mouse is over.and add an option not refresh variation when displayed for mainboard.
And i think we need not only the first variation so i add right/left click on subboard to swtich bettween variations


yzyray:
Add mouseWheel to control branch length in subboard

PR #651, opened 2019-11-01
Add play sound

yzyray:


featurecat:
The sounds are so satisfying! However, it sounds a little bit laggy. For example, if I press the down arrow 3 times so that the next 3 moves are played, the sounds I hear do not correspond to exactly the amount of time I took to press all 3 moves. They sound combined together. Do you have any ideas on how to improve this before I merge?

yzyray:
> The sounds are so satisfying! However, it sounds a little bit laggy. For example, if I press the down arrow 3 times so that the next 3 moves are played, the sounds I hear do not correspond to exactly the amount of time I took to press all 3 moves. They sound combined together. Do you have any ideas on how to improve this before I merge?

I made it on purpose,beacuse i find if play sound on every moves,when we jump many moves fast.the sound effect will be very loud.so i set only play sound one time in 100 millisecond.

featurecat:
Oh. I'll take a look when I have some time to see if I can improve it at all.

PR #656, opened 2019-11-03
Option to opaquely paint grayscale territory prediction on board

TFiFiE:
Following up on https://github.com/featurecat/lizzie/pull/571#issuecomment-515765015. The keyboard shortcut to toggle this setting is **Shift-Period**.


TFiFiE:
Note: `Menu.view.kataGo.kataEstimate.blend` will need translations.

PR #664, opened 2019-11-23
Small sysout println changes

toomasr:
I noticed that lizzie prints out the language code on each startup which can be mysterious for the user. I changed the messaging there so that user would not be that confused about just seeing "en" for example.

Also I've noticed that most of the issues I have is that Lizzie cannot start an engine. I added a debug line to print out the command it is trying to execute. I'm sure this will help many people to debug the configuration of Lizzie.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+.classpath
+.project
+.settings
+gtp.log
 persist.properties
 persist
 lizzie.properties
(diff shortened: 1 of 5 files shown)


toomasr:
Of course now I noticed that it bundled the other pull request contents also into this one. My bad. Just to save time I won't go fixing this yet as maybe the other pull requests will also get in.

PR #546, opened 2019-06-26
Support katago

yzyray:
Support katago, using kata-analyze instead of lz-analyze,show dynamic estimate on board,show scoreMean on suggestion moves and draw scoreMean graph like winrate graph.
kata2

PR #684, opened 2020-02-11
Remove redundant score display to fix #683

### LINKED ISSUE #683: Score is shown as "0.0" in saved analysis of KataGo 1.3
I auto-analyzed a game with KataGo 1.3 and saved the SGF file including the analysis. Then I opened that SGF file again. With "Pondering off", I assume it uses the saved information for each node. See the attached screenshot.

The problem is that while the violet graph (correctly) shows that Black is leading by 27.3 points for the selected move, the number in the top middle always shows "0.0".

Maybe it has to do with how KataGo 1.3 changed the score estimation (see KataGo's release notes for 1.3).

lizzie-saved-points-bug

kaorahi:
These patches revert parts of my past PR #542 (2814463) because #546 (c8ac633) gives a better implementation of the same feature. :-)

We wrote them independently, and the estimated score is doubly shown above and below the winrate bar now...

### DIFF
--- a/src/main/java/featurecat/lizzie/analysis/Leelaz.java
+++ b/src/main/java/featurecat/lizzie/analysis/Leelaz.java
@@ -715,7 +715,6 @@ public class Leelaz {
 
   public class WinrateStats {
     public double maxWinrate;
-    public double maxScoreMean;
     public int totalPlayouts;
 
     public WinrateStats(double maxWinrate, int totalPlayouts) {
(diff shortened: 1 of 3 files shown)


NTUST-MITLAB:
The information is actually different above and below the winrate bar... the above ScoreMean is the overall score lead at the end of the game, the below ScoreMean with the stdev (if set with katago setting of ScoreMean Info with ScoreMean + komi) is the estimate stone difference count on the board which is more useful information in actual counting. 

I know we can just add them manually, but it would be more intuitive just to show them both without the additional mental math to add them together.

kaorahi:
Yes. The displayed scores are different above and below the winrate bar. But their difference is not what you says.

The "below" one is the estimated score of the single best move. The "above" one, that was removed in 0.7.4, was the weighted average of the estimated scores for all suggested moves. Each has its merits and demerits.

hope366:
The screenshot below is Lizzie-v 0.7.3.
The number above the winning percentage bar is a weighted average, but it is strange to show a negative number in this situation.
Is lizzie confused about "+" and "-"?
無題

Issue #676, opened 2020-01-08
Word wrap in the comments box


alexpenson:
Would it be possible to add word wrap in the comments box?
I'm going through commented games on a mac and horizontal scroll of the comments box doesn't go smoothly at all.
Something like this?
https://stackoverflow.com/questions/11300094/jtextpane-dynamic-word-wrap
Thanks!

PR #696, opened 2020-03-23
Commentpane to swing component

toomasr:
I was bored yesterday and I was going through issues in Lizzie and I saw that wordwrap was requested for the comment pane in issue #676 

I jumped into the code and realised that all components are converted into graphics and then those are stitched together for the UI.

As a thought exercise I extracted the comment pane into a separate component, JTextArea. I also had to extract the tree but I still left that as a graphics component.

It could be a step forward to having Java components rather than just graphics.

Another big change in this pull request is I'm skipping a lot of repaints for Leelaz. I saw that the UI sluggish because painting was happening very often when the engine is providing all kind of info. As most of the info you can ignore (it does not matter if the scores change 10 times a second or 2 times) then I'm skipping repaints unless there 700ms has passed since the previous one.

I realise that this is not the best pull request as it is based on my working tree that has other fixes in 
it but I'm still submitting this as the component conversion can be taken separately if there is interest.

The same goes for the UI changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+.classpath
+.project
+.settings
+gtp.log
 persist.properties
 persist
 lizzie.properties
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -271,7 +271,7 @@ public class Config {
 
     gtpConsoleStyle = uiConfig.optString("gtp-console-style", defaultGtpConsoleStyle);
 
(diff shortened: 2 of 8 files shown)


featurecat:
Thanks for the PR. I'll review the changes. Although I'm a bit hesitant about the 700ms updating optimization. Because the engine updates 10 times per second is actually tweakable in the config. Once I take a closer look I'll be able to give better feedback.

toomasr:
Oh, if that is configurable then my patch is definitely not needed. Maybe should just consider different defaults and thats it.

PR #735, opened 2020-07-11
Fix #732 etc. (dynamic-winrate-graph-width)

### LINKED ISSUE #732: Win rate graph bugs
If you load the pre-analyzed Sgf file and proceed from the beginning, if "Automatic width adjustment of winning percentage graph" is enabled, the winning percentage graph will be displayed across the screen as shown in the screenshot below.
lll

kaorahi:
This PR fixes #732 and additional issues (mouse click, score curve) on the winrate graph when dynamic-winrate-graph-width (#384) is enabled.


hope366:
Thank you for your quick response.There is no problem with fixing #732.
* mouse click on the winrate graph for dynamic-winrate-graph-width
* score curve for dynamic-winrate-graph-width
I did not understand the difference between these two items before and after modification.
Please tell me more about these two items.

Issue #734, opened 2020-07-11
Problems or uncertainties when playing a new game with lizzie


hope366:
* Even if you use an engine other than Leela Zero, the player information in the settings dialog and the bottom of the board will always be displayed as "Leea Zero".
* If you set KataGo as the engine and play a new handicap game, the placement of the stones is randomly determined every time →I want an option that allows formal placement by default and free placement as desired.
* Problem at the end. I want you to display the game result when both sides pass. I think it will be easier to understand if the areas of each other are visible.

hope366:
Thank you for correcting the display of "AI" and the initial placement of the handicap game.
Once we have established a way to handle the end of the game, we would like to release it along with some other fixes, are you sure?

hope366:
Is it possible to play engines against each other?
In that case, it is necessary to set the upper limit of the number of searches and the time setting.
The upper limit of the number of searches can be specified with each engine command, but there is no way to specify the time settings individually.
Is it possible to program so that it can be set individually like Sabaki?

PR #736, opened 2020-07-11
Fix obsolete "Leela Zero" for the default player name

kaorahi:
ref. #734


hope366:
Thank you for your support.
There was a flaw in the content of #734. It's just about playing games, but it's actually about handicap games. I fixed #734 as such.

PR #747, opened 2020-08-02
Show error dialog for typical engine troubles

kaorahi:
Many many users are confused by the misleading message "Engine is loading...". It will be better to show an error dialog immediately when Lizzie failed to start the engine, at least.

### DIFF
--- a/src/main/java/featurecat/lizzie/analysis/Leelaz.java
+++ b/src/main/java/featurecat/lizzie/analysis/Leelaz.java
@@ -476,6 +476,9 @@ public class Leelaz {
       // this line will be reached when Leelaz shuts down
       System.out.println("Engine process ended.");
       if (!isQuittingNormally) {
+        if (!Lizzie.gtpConsole.isVisible()) {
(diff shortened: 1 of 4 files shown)


kaorahi:
I added several commits into this PR for #756. See also #759.

kaorahi:
I added another commit so that "Engine is down." is shown immediately (ref. https://github.com/featurecat/lizzie/issues/752#issuecomment-699501390).

This is an example of a wrong weight file (combined with #759).
wrong_weight

hope366:
Oh, I wasn't aware of the big mistake.
* Show "Engine is down." In trouble instead of "Engine is loading ..."(https://github.com/featurecat/lizzie/pull/747/commits/9ed6b840463adc35d123678f79ada2efc5f9c83c) 

LizzieFrame.java in this fix list I forgot to apply the part about 😥
I added it immediately and it worked 😄

This is an example of a mistake in the engine command "gtp"-> "tp".
<div align="center">

PR #763, opened 2020-09-29
Correct inconsistent indents

kaorahi:
It is annoying that `mvn package` modifies indents of GtpConsolePane.java every time. I will be glad if @featurecat merges this PR into master soon for comfortable developments.


kaorahi:
thx!

PR #771, opened 2020-10-04
added maven wrapper

jim-butcher:
Converted to using the maven wrapper. This allows better control of the build chain versions and
makes contributions more consistent.

Currently using the plugin from [takari](https://github.com/takari/maven-wrapper)
which in being brought into apache as of maven 3.7.


jim-butcher:
ref: https://maven.apache.org/plugins/maven-wrapper-plugin/

PR #780, opened 2020-10-12
Fix wrong "Engine is loading..."

kaorahi:
Fix the following issues.

Issue A:
Hit the space key while "Engine is loading..." is displayed. Then this message remains even when loading is finished actually.

Issue B:
Switch the engine from the menu and never move the mouse after the click. Then "Engine is loading..." is never shown until you move the mouse.

### DIFF
--- a/src/main/java/featurecat/lizzie/analysis/EngineManager.java
+++ b/src/main/java/featurecat/lizzie/analysis/EngineManager.java
@@ -129,6 +129,7 @@ public class EngineManager {
     } catch (IOException e) {
       e.printStackTrace();
     }
+    Lizzie.frame.refresh();
   }
 
   public void refresh() throws JSONException, IOException {
--- a/src/main/java/featurecat/lizzie/analysis/Leelaz.java
+++ b/src/main/java/featurecat/lizzie/analysis/Leelaz.java
@@ -344,14 +344,6 @@ public class Leelaz {
       } else if (line.equals("\n")) {
         // End of response
       } else if (line.startsWith("info")) {
-        if (!isLoaded) {
-          Lizzie.frame.refresh();
-        }
-        isLoaded = true;
-        // Clear switching prompt
-        switching = false;
-        // Display engine command in the title
-        Lizzie.frame.updateTitle();
         if (isAnalysisUpToDate()) {
           // This should not be stale data when the command number match
           if (isKataGo) {
@@ -404,6 +396,14 @@ public class Leelaz {
         isThinking = false;
 
       } else if (line.startsWith("=") || line.startsWith("?")) {
+        if (!isLoaded) {
+          isLoaded = true;
+          // Clear switching prompt
+          switching = false;
+          // Display engine command in the title
+          Lizzie.frame.updateTitle();
+          Lizzie.frame.refresh();
+        }
         if (printCommunication || gtpConsole) {
           System.out.print(line);
           Lizzie.gtpConsole.addLine(line);
--- a/src/main/java/featurecat/lizzie/gui/Menu.java
+++ b/src/main/java/featurecat/lizzie/gui/Menu.java
@@ -1247,6 +1247,7 @@ public class Menu extends JMenuBar {
               Lizzie.leelaz.isThinking = false;
             }
             Lizzie.leelaz.togglePonder();
+            Lizzie.frame.refresh();
           }
         });
     analyzeMenu.add(toggleAnalyze);


hope366:
After applying this PR, it seems that lizzie often fails to load.
Since it does not accept any operation and cannot be closed, it will be closed in the task manager.
There is no load on the computer.
It does not occur in normal mode, only in panel UI mode.
Also, it does not occur when the GTP console is set to auto display.
This means that this bug can occur if the GTP console is hidden in panel UI mode.
It seems that the frequency of occurrence increases by pressing the space key several times before the loading is completed.
The same error occurs before applying this PR, but it occurs much less frequently.
I will post a screenshot for the time being, but I don't think there is any particular information available. If the GTP console is set to be displayed at the same time as startup, this error does not occur, so the information on the GTP console is unknown.
無題

NTUST-MITLAB:
There is an associated bug regarding the toggle of "Pondering on" and "Pondering off" under Panel UI mode. The message gets toggled using the spacebar to toggle analyze, but didn't toggle immediately using the menu Analyze -> Toggle analyze(space). Only the green and red indicator dot above the winrate graph updated.

On the other hand, with Panel UI off, the toggle worked for the menu to switch the message.

hope366:
There are two items in this PR, but a simple experiment was conducted to find out the cause of lizzie's failure to start.
lizzie fails to start after applying [Fix forever "Engine is loading ..." if pondering is off](https://github.com/featurecat/lizzie/pull/780/commits/c3a259ced746b0995d49a5c61e5ba01322cff12b) is. There is a 50% or higher chance that it will fail to start.
On the other hand, if you only apply [Fix unshown "Engine is loading ..." until mouse move](https://github.com/featurecat/lizzie/pull/780/commits/f6958db7ec4366106bb70e00dfc6c05c2404bc0c), lizzie will not fail to start. did.
As I mentioned in the previous post, this bug does not occur in normal mode, only in panel UI mode.
Also, if the GTP console is set to be displayed at the time of startup, this problem does not occur.
In other words, this bug occurs when the GTP console is hidden in panel UI mode. And the probability of this bug occurring is very high, 50% or more.

NTUST-MITLAB:
I think some repaint() are not place and triggered in the right place under the Panel UI mode. I noticed that sometimes when I trace a variation on the mainboard, the variation tree doesn't update to the correct node. it stays the same until a new move is made on the mainboard. And sometimes, right-click on the mainboard will pop up the menu for addSuggestionAsBranch here instead of undo a move, even though it is not a candidate location yet.

NTUST-MITLAB:
It stills works on my end for successful starting (didn't break the currently working one), but the issue of clicking on the mainboard, but the variation tree not updated still exists, and the delay is still there.  And using the menu to Toggle Analyze, still not updating the "Pondering on", and "Pondering off" with Panel UI on.

kaorahi:
@NTUST-MITLAB
The pondering status should be updated by the following patch. I will make a PR for it later. I also pushed c3a69a0 into #773. Would you show the steps to reproduce your other issues if possible?

~~~
--- a/src/main/java/featurecat/lizzie/gui/Menu.java
+++ b/src/main/java/featurecat/lizzie/gui/Menu.java
@@ -1247,6 +1247,7 @@ public class Menu extends JMenuBar {
               Lizzie.leelaz.isThinking = false;
             }
             Lizzie.leelaz.togglePonder();
+            Lizzie.frame.refresh();
           }
         });
     analyzeMenu.add(toggleAnalyze);
~~~

hope366:
> And sometimes, right-click on the mainboard will pop up the menu for addSuggestionAsBranch here instead of undo a move, even though it is not a candidate location yet.

I couldn't reproduce it well. Do you mean to right click anywhere other than a candidate move with some stones on the main board?

For the pondering status, please refer to the PR of [Fix: "Pondering on / off" was not updated by Analyze button](https://github.com/featurecat/lizzie/pull/660). Has been fixed, but it seems that the lower left display does not update when switching from on to off from the top menu. The bottom left display doesn't seem to update unless I left-click or move the mouse pointer over the mainboard.

NTUST-MITLAB:
> @NTUST-MITLAB
> The pondering status should be updated by the following patch. I will make a PR for it later. I also pushed [c3a69a0](https://github.com/featurecat/lizzie/commit/c3a69a0ea48c28950628bb68dce88417e7fa6970) into #773. Would you show the steps to reproduce your other issues if possible?
> 
> ```
> --- a/src/main/java/featurecat/lizzie/gui/Menu.java
> +++ b/src/main/java/featurecat/lizzie/gui/Menu.java
> @@ -1247,6 +1247,7 @@ public class Menu extends JMenuBar {
>                Lizzie.leelaz.isThinking = false;
>              }
>              Lizzie.leelaz.togglePonder();
> +            Lizzie.frame.refresh();
>            }
>          });
>      analyzeMenu.add(toggleAnalyze);
> ```

Applying https://github.com/featurecat/lizzie/pull/780/commits/ebb6ea2f7304d08d61f69a9a87660fb5cc81894d fixed the "pondering on/off"...

The other issue of tree variation not updated, seems to occur when there is a wide variations selection exceeding the display at the same time within some variation branches Alt+M were used to change moves from before the variations branched out (this happened sometimes when following live games, when the transcribe reverse previous errors, instead of redoing all the moves, easier to use Alt+M). I think the main issue is that the branches seem to keep different "history moves" compared to the main branch with Alt+M. When I use SetAsMain to switch branches, those history moves seem to flip, and if those moves later on involved in like ko fight or got killed, it messed up the update.

NTUST-MITLAB:
> > And sometimes, right-click on the mainboard will pop up the menu for addSuggestionAsBranch here instead of undo a move, even though it is not a candidate location yet.
> 
> I couldn't reproduce it well. Do you mean to right click anywhere other than a candidate move with some stones on the main board?
> 

I think I found the problem, not particularly related to Iizzie, but more to do with combination with a particular Katago setting. I usually set analysisWideRootNoise to a very high value in order to discover unusual moves Katago sometimes misses. When this value is high enough but not too high, hence it spread to other nodes even remote ones passing a certain threshold after a while (not immediately), and I was investigating the right-click/left-click delay issue, so I set the Update Interval to very short (1). The combination sometimes messes up the refreshing when I right-click and left-click just at the right moment, where I'll see a ghost image of both black and white stone on the same spot, and that in turn mess up the menu pop up. 

This was not an issue before if when update interval is a normal value like 50 or 100, it seems there is a lower bound to how fast it can be set.

hope366:
It's hard to replicate the problem.
What exactly is the value of analysisWideRootNoise set to?
It would be helpful if you could provide a specific diagram of the aspects of the malfunction.

NTUST-MITLAB:
> It's hard to replicate the problem.
> What exactly is the value of analysisWideRootNoise set to?
> It would be helpful if you could provide a specific diagram of the aspects of the malfunction.

I set the analysisWideRootNoise to 0.2. 

But as I tried different value, I found that as long as it is large enough to spread possible next move further, the higher its value, the more chance this would happen when right-click/left-click quickly on different locations of the board (usually during mid-game, where lots of possible locations could be possible next moves). 
addSUggestionAsBranch bug

hope366:
I set analysisWideRootNoise to 0.2, made numerous branches, and repeated the right and left clicks frequently, and apparently I was able to reproduce it.
It seems to have happened when the candidate moves are displayed in many places on the board.
I think I'm probably referring to the screenshot below (Japanese version)
However, I found it difficult to find an absolute condition that would always occur in these cases.
無題

NTUST-MITLAB:
> I set analysisWideRootNoise to 0.2, made numerous branches, and repeated the right and left clicks frequently, and apparently I was able to reproduce it.
> It seems to have happened when the candidate moves are displayed in many places on the board.
> I think I'm probably referring to the screenshot below (Japanese version)
> However, I found it difficult to find an absolute condition that would always occur in these cases.

That's why I said "sometimes" and "click at the right moment" previously. Using Lizzie to review a live game, players might try many different moves and investigate deep in these variations, but normally it would take a moment to think about the next move and things work out fine. 

However, sometimes there are obvious replies has to make with different following sequences when you don't have to think (like in capturing race, ladders, net, etc.), then under that condition, the quick "left-click" and "right-click" to go back to different branch points would come up a lot. That's why I think during mid-game where there are many candidate moves and far apart, this bug is most likely to happen "naturally". And I think this has something to do with how the variation tree keeps variations behind the visible circle display.

NTUST-MITLAB:
I got a clue about the addSuggestionAsBranch problem. Whenever this occurred, the right-click action triggers commands to send both kata-analyze and name command to the gtp console at once, effectively start, and immediately stop the analysis engine and the addSuggestionAsBranch pop-up.  And when right-click again somewhere to exit the addSuggestionAsBranch, only the analyze command is sent again. And there is a small delay for the first variation to be drawn on the main board, and another delay for all the different color circles to be drawn, before the right-click resume the function of undo.

Issue #777, opened 2020-10-11
Lizzie cannot remember Panel -> GtpConsole detached window is checked


NTUST-MITLAB:
Normally I'd detached the Gtp Console so it can be resized and moved to a convenient location as a separate window. 

After updating to 0.7.4, Lizzie cannot remember it has been opened, and have to check View -> Panel -> Gtp -> GtpConsole (or use hotkey alt+E) every time opening Lizzie.

hope366:
There are the following eight types of "panel display", but only "status" and "Toolbar" are memorized for changes, and others are not memorized.
・Sub board
・Win rate graph
・comment
・Change chart graph
・Agehama
・status
・GTP console
・Toolbar
I think it would be more convenient if lizzie remembers "display ⇔ hide" and the size of the window for all these 8 items.
In addition to this, you can now switch the scaling of the sub board by clicking the center, but it would be convenient if this function could also be used in the winning percentage graph.
It may be useful for lizzie to remember about resizing with a center click as well.

NTUST-MITLAB:
> There are the following eight types of "panel display", but only "status" and "Toolbar" are memorized for changes, and others are not memorized.
> ・Sub board
> ・Win rate graph
> ・comment
> ・Change chart graph
> ・Agehama
> ・status
> ・GTP console
> ・Toolbar
> I think it would be more convenient if lizzie remembers "display ⇔ hide" and the size of the window for all these 8 items.
> In addition to this, you can now switch the scaling of the sub board by clicking the center, but it would be convenient if this function could also be used in the winning percentage graph.
> It may be useful for lizzie to remember about resizing with a center click as well.

Gtp Panel display used to be default true (IIRC before 0.7x?) when print_comms config can still be set, but was turned off to speed up the loading, and again a toggle function is added, and it would be toggled on showing the Tuning initialization instead of just a long "Engine Loading" waiting. Also when the Gtp Concole window reopened, Lizzie still remembers its previous location and resizing, hence some of the information of the window is kept already. All the components are there to make it a full-on panel (maybe an additional flag? and make Gtp Console visible at the right moment). Sometimes I can trick it to "keep the window" when Lizzie was hanged, and I process killed it without normal exit, which I believe bypasses the toggle in some earlier version.

The Panel UI with LizzieMain can have more improved layout management and make customization better and more intuitive.

kaorahi:
[...]

PR #782, opened 2020-10-15
Hopefully fix hang-up that was reported in #777 and #780

kaorahi:
Ensure some doubtful methods to be called in EDT.

ref.
https://github.com/featurecat/lizzie/issues/777#issuecomment-706775010
https://github.com/featurecat/lizzie/pull/780#issuecomment-707220287

### DIFF
--- a/src/main/java/featurecat/lizzie/Lizzie.java
+++ b/src/main/java/featurecat/lizzie/Lizzie.java
@@ -10,6 +10,7 @@ import featurecat.lizzie.rules.Board;
 import java.io.File;
 import java.io.IOException;
 import javax.swing.JOptionPane;
+import javax.swing.SwingUtilities;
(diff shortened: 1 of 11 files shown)


hope366:
I just applied the second PR.
The first PR alone seemed to solve the problem, but does the second PR give you even more stability?
In any case, thank you for demonstrating excellent PR.

kaorahi:
In my environment, Lizzie shows an empty window instead of the alert dialog "Lizzie requires version 0.15 or later of Leela Zero for analysis" if I set `leela_gtp -g` as the default engine after a6280681. c500a89d fixes this.

kaorahi:
I pushed 2ce1d7f4 to fix kaorahi#3 (flicker of ownership colors in auto-analysis), that occurs after d790275e.

Issue #785, opened 2020-10-15
About the display method when switching engines


hope366:
You can see the name of the weight file in the pull-down menu on the top menu to switch engines.
When the same weight file "KataGo40b" is used with different engines, KataGo-v1.6.1-OpenCL and KataGo-v1.6.1-CUDA, the pull-down menu shows exactly the same name, "KataGo40b", which is very difficult to distinguish. It is.
It's possible to have two files that are identical and have different filenames, like "KataGo40b-1.6.1-OpenCL" and "KataGo40b-1.6.1-CUDA", but that's a bit tricky.
It would be useful if you decide on an arbitrary nickname that is easy to understand and the nickname appears in the pull-down menu.
Sabaki uses this technique, and I find it useful.

NTUST-MITLAB:
Without a rework and adding additional fields for aliases to different engine settings (it would involve saving alias, manu item layout arrangement, even submenus), the easiest way to tell the difference as a temporary fix is to change the   `public void updateEngineMenu(List<Leelaz> engineList) ` in Menu.java to show the full engine commands. Instead of appending `engineDt.currentWeight()` to each item, appending `engineDt.engineCommand()`

hope366:
Thank you for showing me how to make useful changes.

![無題](https://user-images.githubusercontent.com/63999713/96331593-7b52d980-1099-11eb-88c3-1ddf5c17db0f.png)
I think it's a great idea in the sense that it's possible to completely distinguish between the different engines.
On the other hand, the display content may be long and somewhat difficult to read visually.
It's also a shame that you can't add your own favorite information such as "analysisWideRootNoise = 0.0" or "rules = japanese".

hope366:
Thank you for your prompt response as usual.
![無題](https://user-images.githubusercontent.com/63999713/96368478-f8ac4600-118e-11eb-9be3-9d4c89e883aa.png)

PR #787, opened 2020-10-17
Quick hack to support #785 (<nickname> before the engine command)

kaorahi:
You can write `<foo> ./leelaz -g -w network.gz` in the menu Settings > Engine for example. Then `foo` is used as the label in the engine menu and `./leelaz -g -w network.gz` is used as the actual engine command.

This is just a quick hack to support #785. It is not a good way. But it seems to be desired strongly by some people ([ref.](https://github.com/featurecat/lizzie/issues/671#issuecomment-640103271)).


NTUST-MITLAB:
As long as the nickname doesn't have a `<` character inside... just a fair warning, pattern matcher is sometimes dangerous...

PR #795, opened 2020-10-20
Fix #786 (Problems with saving and loading files)

### LINKED ISSUE #786: Problems with saving files
When saving a file in lizzie, `*.sgf` is specified by default.
If you save the file in this state, it will be saved as a `.sgf` file, but when you save another new file, you won't be able to see previously saved files of the `.sgf` format.
The `*` doesn't seem to be necessary. What is the purpose of `*`?

hope366:
Fixed an issue where existing files were not displayed when saving the file.
Fixed an issue where uppercase .SGF or .GIB files could not be loaded.

This PR is based on @kaorahi's hint and @NTUST-MITLAB's idea.

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/MainFrame.java
+++ b/src/main/java/featurecat/lizzie/gui/MainFrame.java
@@ -343,17 +343,13 @@ public abstract class MainFrame extends JFrame {
         || file.getPath().endsWith(".gib")
         || file.getPath().endsWith(".SGF")
         || file.getPath().endsWith(".GIB"))) {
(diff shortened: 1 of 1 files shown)


NTUST-MITLAB (changes requested):
Maybe we should remove the debug messages of System.out.println.

hope366:
@NTUST-MITLAB Thank you for the advice. I wasn't sure if I should delete the debugging message or not, but I certainly found it unnatural to keep it.
Thank you very much for your kindness in addressing my question in Issues.

PR #802, opened 2020-10-30
Fix #423 (node blunder colors in the default theme)

### LINKED ISSUE #423: Node blunder colors don't work in the default theme
@zsalch 

For now, it's not a big problem. It can be an optional feature, especially since I think the colors need to be chosen more carefully. I think a gradient from green to neutral to red would be best. Green for "good moves", red for "bad moves"

kaorahi:
I'm afraid most users may not notice the feature #413 because of the issue #423. This PR applies the proposed parameters in https://github.com/featurecat/lizzie/issues/423#issuecomment-438878060 if they are empty. It works only once so that users can still set them empty again if they like.

related: #800 #801

### DIFF
--- a/src/main/java/featurecat/lizzie/Config.java
+++ b/src/main/java/featurecat/lizzie/Config.java
@@ -199,16 +199,14 @@ public class Config {
     if (alreadyTriedOnce) return false;
 
     boolean modified = false;
-    String theme = ui.optString("theme", "");
     JSONArray blunderWinrateThresholds = ui.optJSONArray("blunder-winrate-thresholds");
     JSONArray blunderNodeColors = ui.optJSONArray("blunder-node-colors");
-    boolean isDefaultTheme = theme.toLowerCase().equals("default");
     boolean isEmptyBlunderWinrateThresholds =
         (blunderWinrateThresholds == null || blunderWinrateThresholds.length() == 0);
     boolean isEmptyBlunderNodeColors =
         (blunderNodeColors == null || blunderNodeColors.length() == 0);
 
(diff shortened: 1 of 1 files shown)


hope366:
I have been using lizzie with this PR for quite some time.
Right now, when I put that lizzie in the official lizzie 0.7.4 folder and started it, the winning percentage change node was left blank.
In the official lizzi 0.7.4, the winning percentage change node is blank by default, so I thought this PR would work. Is this recognition wrong?

kaorahi:
thx. It seems broken for some reason. Please apply 64dc6291 to fix it.

When you test it, you need to delete (or rename) both files `config.txt` and `persist`.

hope366:
I tried the modified version and it seems to work fine. Thank you very much.
When I started lizzie after deleting "config.txt", an error occurred because the part of the engine command that should be "./leelazero/leelaz.exe" is "./leelaz". It worked normally by rewriting it correctly. 
無題

kaorahi:
I found the above "some reason". The name of the default theme is inconsistent between Config.java ("default") and ConfigDialog.java ("Default" in English, or i18n names in other languages).

So 64dc6291 can fail in non-English environments. The correct fix should be cd14ee86. Namely, we do not need to check the default theme here because we never touch `theme/yasnaya/theme.txt` anyway in `checkEmptyBlunderThresholds` from `validateAndCorrectSettings`.

cd14ee86 should work when all the following conditions are satisfied.

1. "blunder-winrate-thresholds" and "blunder-node-colors" are both empty in `config.txt`.
2. "checked-empty-blunder-thresholds" is not true in `persist`.

PR #804, opened 2020-11-01
Fix #781 (slow play/undo/redo)

### LINKED ISSUE #781: Slow play/undo/redo with Panel UI + ToolBar + KataGo
Undo takes about 0.5 sec if I use Panel UI + ToolBar + KataGo (OpenCl). It is smooth otherwise. Similar for play and redo. This issue occurs even if I delete most of ToolBar.java as follows.

~~~
package featurecat.lizzie.gui;

import javax.swing.JToolBar;

public class ToolBar extends JToolBar {
  public ToolBar() {}
}
~~~

Lizzie 0.7.4, KataGo 1.6.1, Debian GNU/Linux stable (buster), Intel Core i7-8550U, no GPU

kaorahi:
In Panel UI, the layouts of all the UI parts are rearranged in every play/undo/redo after d2c587cdb. #781 seems to be fixed if we stop this redundant work. Can anybody test this PR for a week or so and report whether you notice any problem?

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/LizzieMain.java
+++ b/src/main/java/featurecat/lizzie/gui/LizzieMain.java
@@ -542,7 +542,6 @@ public class LizzieMain extends MainFrame {
     commentPane.drawComment();
     //    commentPane.revalidate();
     commentPane.repaint();
-    invalidLayout();
   }
 
   @Override
--- a/src/main/java/featurecat/lizzie/gui/LizzieMain.java
+++ b/src/main/java/featurecat/lizzie/gui/LizzieMain.java
@@ -122,6 +122,11 @@ public class LizzieMain extends MainFrame {
               }
               // draw the image
               Graphics2D bsGraphics = (Graphics2D) g; // bs.getDrawGraphics();
+              // Note: #781 is caused by repaint of other parts (toolBar in particular) with
+              // VALUE_ANTIALIAS_ON. To prevent such unwilling side effects, we change
+              // RenderingHints only temporary here and restore their old values at the bottom of
+              // this method.
+              RenderingHints oldHints = bsGraphics.getRenderingHints();
               bsGraphics.setRenderingHint(
                   RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY);
               bsGraphics.setRenderingHint(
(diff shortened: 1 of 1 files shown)


hope366:
I'd like to verify the PR, but I can't reproduce [#781](https://github.com/featurecat/lizzie/issues/781), probably because of the GPU.would it be difficult to do so without a CPU-only PC?
I tried katago-v1.6.1-cpu-eigen.
The GPU doesn't seem to work, but undo was fast.

kaorahi:
thx! You don't need to reproduce the issue #781, that may depend on the environments.

I'm worried about unexpected side effects of the PR. If you notice a problem like broken layouts etc., please let me know the steps to reproduce it.

hope366:
In my environment, applying [#804](https://github.com/featurecat/lizzie/pull/804/commits/1b7761d79283f289b486b509c0052fbd8aa93735) hasn't done anything different I'll continue to monitor it carefully.

By the way, I found a minor issue with the panel UI mode.
Press Alt+W. This will hide the comments panel. You can detach the panel by dragging the mouse, but if you revert the detached panel, the comments panel will remain hidden.
When I restarted lizzie, it was back to normal, so it's not a very serious problem.

NTUST-MITLAB:
I will start a live test during today's (and hopefully tomorrow's) Samsung Fire Cup final match, and see how it fairs.

kaorahi:
@hope366
You are still in the design mode. Hit Alt+W again to exit.

kaorahi:
1b7761d7 is unacceptable because it prevents the message "Engine is loading...". So I reverted it and pushed another workaround. How about this one? Do you notice any change of the graphics quality?

hope366:
>You are still in the design mode. Hit Alt+W again to exit.

Okay, that worked out well.
Alt+W toggles only the comments panel to show/hide, but it's not something that I find particularly inconvenient.

>Do you notice any change of the graphics quality?

I couldn't really see the difference.
I'm curious to see where the difference in quality lies.

NTUST-MITLAB:
During the live game today the old fix did seem to make it faster but has the side effect listed. I'll try another live game sometime with the new fix. And here is today's live game save file, with lots of branches in the variation tree, and all the best moves embedded, maybe others can see if adding new nodes and undo in these big branches would cause lag...
[The 25th Samsung Fire Cup Match 2 - Ke Jei (B) vs Shin Jinseo (W).zip](https://github.com/featurecat/lizzie/files/5482990/The.25th.Samsung.Fire.Cup.Match.2.-.Ke.Jei.B.vs.Shin.Jinseo.W.zip)

kaorahi:
thx for your reports. Though 28f2c33b accelerates the response of Lizzie remarkably in my environment, I'm afraid of unexpected side effects like degraded graphics. So if you notice nothing, it is a good news for this PR.

PR #807, opened 2020-11-04
Fix #805(The ">>" doesn't work properly)

### LINKED ISSUE #805: The toolbar Fast forward 10 moves doesn't work properly
The fast forward 10 moves button `>>` on the toolbar ` |<  << < > >> >|` only forward 1 move, instead of 10 (although the rewind 10 moves `<<` does work).

hope366:
fix #805  
The ">>" on the lower toolbar is originally a function that "advance 10 moves", but there was a bug that only one move was actually made.
It seems that there was a mistake in specifying the number, so I corrected it.


NTUST-MITLAB:
I think this is just a typo error and should be fixed by this.

hope366:
Thank you.

Issue #810, opened 2020-11-09
Wrong score in winrate graph when it is missing


kaorahi:
1. Start Leela Zero on Lizzie 0.7.4.
2. Analyze a game and hit END key.
3. Switch the engine to KataGo and hit SPACE key to stop pondering.
4. Hit the up arrow key.

Then we see a wrong score 0.0 in the winrate graph.

In Panel UI, we also get #729 again because I made a wrong fix of the conflict in the merge dc8f872f. The line "validScore = " should be placed BEFORE "if (!validWinrate)" in WinratePane.java.

hope366:
Leela Zero doesn't analyze Score Mean, so I think it's natural to show 0.0, but am I doing something wrong?
Also, as I experimented with this issue, I found another one.

1. After launching lizzie 0.7.4, press the spacebar to turn off pondering.
2. Next, arrange about 10 stones on the main board.
3. Press the spacebar to turn on pondering.
4. Press the up arrow key to return the main board to an empty state.

At this time, stdev and Lastmove are not displayed until the move number becomes 1.
If you turn off pondering and place stones with the down arrow key, they will be displayed, so it seems that the calculation was done internally.

kaorahi:
I expect blank when pondering is off and no estimation is available. But 0.0 is displayed actually in my previous post https://github.com/featurecat/lizzie/issues/810#issue-739024656 .

hope366:
After all I seem to have misunderstood a little.
Whether it's LZ or KataGo, if you line up the stones with pondering off, no analysis results will be displayed.
If ScoreMean shows 0 by changing the engine to KataGo after analyzing with LZ, does it mean that there is an instruction to set ScoreMean = 0 at the time of analysis with LZ?

Issue #808, opened 2020-11-06
About the display of the winning percentage graph


hope366:
Currently, there is a display "mean: ### stdev: ### Last move: ###" at the top of the winning percentage graph.
When pondering off, only "Last move: ###" is displayed.(Normal mode only)
How about adding "current turn" and "win percentage number" to the information here?
In particular, I think it is easier to understand if there is a display of "current turn". I think stdev and Last move have low priority, so I think it's okay to hide them in some cases.
For example, "Turn: black Winrate: 50.64% Score Lead: 0.1"
How about something like this?

See the screenshot below.
This is the information in the white turn. Although it is "mean: 5.4", the numerical value in the graph is "-5.4". The numbers in the graph are always the numbers seen from the black side, but when the turn is white, it is easier to understand if the numbers seen from the white side are displayed.
Next, at the bottom of the graph, there is a number indicating the number of steps., but it is difficult to see due to the green line. It is a small matter which one to prioritize, but I feel that it is more natural to prioritize the number of steps display.
<img src="https://user-images.githubusercontent.com/63999713/98323196-9e373480-202c-11eb-995e-74dad77ee5bf.png" width=50%>

NTUST-MITLAB:
> See the screenshot below.
> This is the information in the white turn. Although it is "mean: 5.4", the numerical value in the graph is "-5.4". The numbers in the graph are always the numbers seen from the black side, but when the turn is white, it is easier to understand if the numbers seen from the white side are displayed.

You can change this in `View -> KataGo settings -> ScoreMean View -> Always black/alternately `

hope366:
See the two screenshots below.
Both use exactly the same analysis results in exactly the same aspect.
Now White's turn, White is 5.1 points ahead.
The left is "alternate" and the right is "always black".
<img src="https://user-images.githubusercontent.com/63999713/98378859-ed5f8280-2089-11eb-8f96-2ddcf516cfd8.png" width=40%> <img src="https://user-images.githubusercontent.com/63999713/98378869-f18ba000-2089-11eb-9a34-c14a8fd27523.png" width=40%>
On the left, the problem is that the number on the graph is "5.1" but the number in the graph is "-5.1".
Both on the right show the same number "-5.1", but the problem is that it is a negative number even though White is leading in White's turn.

NTUST-MITLAB:
[...]

PR #811, opened 2020-11-14
Fix some of the issues in #808 and #810 (score text etc.)

kaorahi:
See the commit logs for details.


hope366:
>Fix: score is not shown when the previous move is not evaluated (#810)

Is this PR a fix to improve # 810?
I tried to verify, but when I changed the engine to KataGo after analyzing with LZ, the number on the bar graph and the number in the winning percentage graph are 0 for Score Mean.
What is the specific effect of this PR?

kaorahi:
c473734c fixes your "another one" in https://github.com/featurecat/lizzie/issues/810#issuecomment-724102807 . :)

hope366:
I see, it certainly seems to be improving.
It seems that my reading comprehension was a little lacking.
Thank you for creating the PR.

PR #819, opened 2020-11-25
Couple of small fixes 001

toomasr:
As I was analysing a game after a long time with KataGo I noticed that one of the settings isn't saved. There was a saving functionality missing in one of the listeners and then I added that. This is this commit - https://github.com/featurecat/lizzie/commit/a38b127bc8de77ecd1bcba3979fc4675805baeb2

On top of that I thought I'll also cherry pick some small useful fixes whether for users or for developers that are small enough for a quick review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ target
 leelaz_opencl_tuning
 *.sh
 *.jar
+gtp_logs
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+.classpath
+.project
+.settings
+gtp.log
 persist.properties
 persist
 lizzie.properties
--- a/src/main/java/featurecat/lizzie/Lizzie.java
+++ b/src/main/java/featurecat/lizzie/Lizzie.java
@@ -38,15 +38,16 @@ public class Lizzie {
   public static void initializeEngineManager() {
     try {
       engineManager = new EngineManager(config);
-      if (mainArgs.length == 1) {
-        frame.loadFile(new File(mainArgs[0]));
-      } else if (config.config.getJSONObject("ui").getBoolean("resume-previous-game")) {
-        board.resumePreviousGame();
-      }
     } catch (IOException e) {
       frame.openConfigDialog();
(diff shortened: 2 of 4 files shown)


kaorahi:
Hi, @toomasr san. Thanks for your fixes.

I noticed that ConfigDialog.java was modified after `mvn package`. This is very inconvenient in future developments. Though I am not an official staff of this project at all, can I ask you to keep the standard style of the project? (Or, it would be better to delete the empty `else` and these comment lines.)

~~~
diff --git a/src/main/java/featurecat/lizzie/gui/ConfigDialog.java b/src/main/java/featurecat/lizzie/gui/ConfigDialog.java
index 8f53bed1..dc598f7e 100644
--- a/src/main/java/featurecat/lizzie/gui/ConfigDialog.java
+++ b/src/main/java/featurecat/lizzie/gui/ConfigDialog.java
@@ -1830,7 +1830,7 @@ public class ConfigDialog extends JDialog {
               resourceBundle.getString("LizzieConfig.title.engine"), "exe", "bat", "sh");
       chooser.setFileFilter(filter);
     } else {
-      //setVisible(false);
+      // setVisible(false);
     }
     chooser.setMultiSelectionEnabled(false);
     chooser.setDialogTitle(resourceBundle.getString("LizzieConfig.prompt.selectEngine"));
@@ -1931,9 +1931,9 @@ public class ConfigDialog extends JDialog {
       Lizzie.initializeEngineManager();
     }
     try {
-      //if (Lizzie.engineManager != null) {
-        Lizzie.engineManager.refresh();
-      //}
+      // if (Lizzie.engineManager != null) {
+      Lizzie.engineManager.refresh();
+      // }
     } catch (JSONException e) {
       e.printStackTrace();
     } catch (IOException e) {
~~~

toomasr:
Thanks for the feedback. I've removed the dead lines.

toomasr:
@kaorahi what is the easiest way to contact you about general discussion on go software?

kaorahi:
How about using [issues of my GUI](https://github.com/kaorahi/lizgoban/issues) if it is not related to Lizzie?

PR #821, opened 2020-11-28
Fix #820 (playouts "1.5k" is wrongly read as 15000 in SGF)

### LINKED ISSUE #820: Problems when adding analysis to pre-analyzed SGF
I analyzed the SGF file using lizzie.
We analyzed from the first move to the last move. The average amount of analysis is around 500 playouts. Since it is about 500 visits / s, it takes about 1 second for each phase. I saved this.
Next, I wanted to add the amount of analysis and save it by overwriting.
Load the saved SGF and add the analysis.
At this time, I noticed something strange.
In one situation, the value on the board changes about 1 second after the analysis is turned on, but in another situation, it takes more than 10 seconds until the value on the board changes.

kaorahi:
Fix #820. Note that you also need to apply #794.


hope366:
Thank you for visiting #820 and writing the code.
It looks like any corrections or improvements are possible to you!

Issue #792, opened 2020-10-19
Character Encoding


NTUST-MITLAB:
Ever since the patch for this https://github.com/featurecat/lizzie/pull/635, I've always had issues with random player names displayed on top, since it forced "WINDOWS-1252" encoding to "gb2312" which only works with Simplified Chinese, where a system with Traditional Chinese uses "Big5" (which is also WINDOWS-1252). In this case, when the information is changed, it would display with the wrong characters, even if it is saved and load by Lizzie itself.

I've applied a dirty fix myself ever since by adding an encoding method to the `save` function in the `SGFParser.java` with `Writer writer = new OutputStreamWriter(new FileOutputStream(filename), "UTF-8")`  so I can manually change the info and write back as UTF-8 encoding, which would work well even with other platforms. But it is not a good fix by any measure.

I feel it is time to properly add the CA property tag for the encoding in the SGF format like most current sgf software already did (like Sabaki, MultiGo, GoGUI, etc.) to permanently solve the character encoding and display issue disregarding which platform it is in.

PR #823, opened 2020-11-29
Force UTF-8 for saving SGF to fix the second issue in #820

kaorahi:
To "fix" the second issue in #820 with the minimal coding, this PR forces UTF-8 for saving SGF. Is this "solution" acceptable?

(Of course #792 will be better if anyone implements it.)


hope366:
Thank you for writing the code.
I applied PR and verified it, but it seems that the same problem as last time has occurred.
http://www.kihuu.net/sgf/k00000130658.sgf
When this is loaded into lizzie, "星陣圍棋" is displayed under the board, but once saved and then loaded, "□□□ w □□□□" is displayed.

Issue #265, opened 2018-06-01
NullPointerException when attempting to open an SGF


colinwd:
Hey there! First off, thanks for all your hard work on Lizzie - it's awesome!

The latest release (Lizzie.0.5.Windows.x64.GPU) is giving me some trouble when trying to open an SGF. Here's the stacktrace:

```Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
        at featurecat.lizzie.gui.LizzieFrame.openSgf(LizzieFrame.java:211)
        at featurecat.lizzie.gui.Input.keyPressed(Input.java:267)
        at java.desktop/java.awt.Component.processKeyEvent(Unknown Source)
        at java.desktop/java.awt.Component.processEvent(Unknown Source)
        at java.desktop/java.awt.Container.processEvent(Unknown Source)
        at java.desktop/java.awt.Window.processEvent(Unknown Source)
        at java.desktop/java.awt.Component.dispatchEventImpl(Unknown Source)
        at java.desktop/java.awt.Container.dispatchEventImpl(Unknown Source)
        at java.desktop/java.awt.Window.dispatchEventImpl(Unknown Source)
        at java.desktop/java.awt.Component.dispatchEvent(Unknown Source)
        at java.desktop/java.awt.KeyboardFocusManager.redispatchEvent(Unknown Source)
        at java.desktop/java.awt.DefaultKeyboardFocusManager.dispatchKeyEvent(Unknown Source)
        at java.desktop/java.awt.DefaultKeyboardFocusManager.preDispatchKeyEvent(Unknown Source)
        at java.desktop/java.awt.DefaultKeyboardFocusManager.typeAheadAssertions(Unknown Source)
        at java.desktop/java.awt.DefaultKeyboardFocusManager.dispatchEvent(Unknown Source)
        at java.desktop/java.awt.Component.dispatchEventImpl(Unknown Source)
        at java.desktop/java.awt.Container.dispatchEventImpl(Unknown Source)
        at java.desktop/java.awt.Window.dispatchEventImpl(Unknown Source)
        at java.desktop/java.awt.Component.dispatchEvent(Unknown Source)
        at java.desktop/java.awt.EventQueue.dispatchEventImpl(Unknown Source)
        at java.desktop/java.awt.EventQueue.access$600(Unknown Source)
        at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
        at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
[...]

PR #824, opened 2020-12-05
only update last-folder if there is something to update, fixes #753

### LINKED ISSUE #753: Can't open/close SGF on lizzie version in Ubuntu repository
I get "org.json.JSONException: No value for last-folder"

The persist file is in `~/.local/share/leela-zero/` for some reason and it looks like this when it gets created:

```
{
  "autosave":, "" // there's a basic SGF here
  "ui-persist": {},
  "filesystem": {
    "last-folder": ""
  }
} 
```

And there's no newline at the end of the file, in case that's relevant.

I've tried manually editing the value of `last-folder` to various things but the result is always this error.

infinity0:


infinity0:
I am not sure how other people have seen this behaviour, but for me open/save is completely broken without this PR, due to lizzie trying to load the autosave upon startup. This causes `null` to be written to "last-folder" because the autosave file has no parent, which the JSON API treats as deleting that key completely, causing subsequent loads/saves to file.

infinity0:
This also sounds like #265, merging this should fix that too.

PR #827, opened 2020-12-08
Fix wrong getCoord which leads analysis hangs when enable katago ownership in rectangle board

yzyray:


kaorahi:
@yzyray Would you consider merging https://github.com/yzyray/lizzie/pull/1 ?

yzyray:
> @yzyray Would you consider merging [yzyray#1](https://github.com/yzyray/lizzie/pull/1) ?

Thanks for the fix,merged.

Issue #828, opened 2020-12-13
Set region of interest like KaTrain 1.7


kaorahi:
KaTrain 1.7 has a new nice feature "Set region of interest". As a quick experiment, I imported it to Lizzie with VERY BAD PROGRAMMING STYLE. We need overall rewriting if we will support it officially. But I don't have enough energy now.

~~It doesn't work in Panel UI at present.~~

Usage:

* Alt+drag to set region of interest.
* Alt+click to reset it.

https://github.com/kaorahi/lizzie/tree/allow0

kaorahi:
We may prefer `untilDepth = 1` like the released version of KaTrain 1.7 rather than 9999.

hope366:
The screenshot on the left is `untilDepth = 9999` and the screenshot on the right is`untilDepth = 1`.
Does "untilDepth" mean something like this?
Is `untilDepth = 9999` more useful if you want to solve "Tsumego" where only partial answers make sense?
<img src="https://user-images.githubusercontent.com/63999713/102689307-6b45aa80-4240-11eb-8c29-fd9008ed314f.png" width="40%">       <img src="https://user-images.githubusercontent.com/63999713/102689310-700a5e80-4240-11eb-83c4-7f83b99b33aa.png" width="40%">

kaorahi:
For example, if we set `untilDepth = 3`, the restricted region is only applied until the third move. I'm afraid `untilDepth = 9999` may cause unreasonable variations. One can never capture stones by a ladder (Shicho)!

BTW, it also works in PanelUI now.

hope366:
It is difficult to grasp how applying up to the nth movement is expressed on the board.
`untilDepth = 9999`,` untilDepth = 1`
The difference between the two, as far as the screenshots are concerned, is whether the change diagram on the board is confined to a restricted area.
Does `untilDepth = 1` apply to the first move? It's hard to see how this will be shown on the board.
>BTW, it also works in PanelUI now.

Thank you.

youjh72:
Thanks you for your amazing effort. I am wondering about what is the difference between 'PanelUI' and normal.
I didn't know the difference when I checked the Panel UI in Lizzie settings.
Thanks in advance.

hope366:
In panel UI mode, you can separate each panel and customize its size and position individually.
Press "Alt + W" to enter design mode.
You can separate the panels by dragging each panel with the mouse in design mode.
To restore it, click the "x" at the top right of each panel.
[...]

PR #833, opened 2020-12-25
Region of interest

kaorahi:
This is a better implementation of #828, "Set region of interest like KaTrain 1.7".

* Alt+drag to set region of interest.
* Alt+click to reset it.

### DIFF
--- a/src/main/java/featurecat/lizzie/analysis/Leelaz.java
+++ b/src/main/java/featurecat/lizzie/analysis/Leelaz.java
@@ -715,6 +715,7 @@ public class Leelaz {
       isPondering = true;
       startPonderTime = System.currentTimeMillis();
     }
+    Lizzie.board.regionOfInterest.reset();
     sendCommand(
         String.format(
             "lz-analyze %d %s",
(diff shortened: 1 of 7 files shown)


kaorahi:
Fixed a very stupid mistake.

hope366:
I'm using [allow0](https://github.com/featurecat/lizzie/compare/master...kaorahi:allow0), so around line 190 of BoardDate.java looks like the screenshot below It has become.
That is, there is no `if (! Lizzie.board.regionOfInterest.isEnabled ()) return;`.
In this case, what changes should I make to solve the winning percentage graph problem?

yzyray:
maybe need fix #858

kaorahi:
Imported yzyray@a7bee3b6 to fix #858 (unintentional drag). thx!

hope366:
I tried 6bead84, but the bug that the winning percentage graph does not work has not been fixed ...
Just in case, I downloaded https://github.com/kaorahi/lizzie/tree/allow1, compiled and tried it, but it seems that the bug of the winning percentage graph is not fixed like 6bead84.

kaorahi:
I disabled updating of the winrate graph in the restricted analysis to avoid confusion with the full analysis. This is not a bug but the intentional behavior.

BTW, I recommend to revert allow0 branch (#828) and use allow1 branch (#833) instead for the future. I won't support allow0 anymore.

hope366:
>I disabled updating of the winrate graph in the restricted analysis to avoid confusion with the full analysis. This is not a bug but the intentional behavior.

OK.
It seemed that allow_1 was used in Pached 0.7.4, so I switched to allow_1.

PR #848, opened 2021-01-22
Variation windows movenum

xiaoyifang:


hope366:
Thank you for adding a convenient function.
I've always felt uncomfortable with the specifications where the move number starts with "1" when I created the branch.
With the addition of this function, it is easier to visually check the difference between the move numbers on the main board and the variation tree, and I think it makes us feel even more uncomfortable.
Even if you create a branch, it would be easier to understand if it is displayed as "1, 2, 3, 4, 5 ...".
I can't program, so if you can adopt this idea, I'd appreciate it if you could write the code in the meantime.
無題
On this issue, MultiGo has the same style as the current lizzie.
Does this mean it's common and makes sense?

xiaoyifang:
code has changed accordingly,check out

hope366:
thank you very much😄

hope366:
When loading an SGF with many branches, it seems to have the glitches shown in the screenshot below.

無題

hope366:
The number of branches doesn't seem to matter.
If you save it in the state on the left and load the saved SGF, it will look like the one on the right.

xiaoyifang:
seems like a bug in save sgf function.when i have time ,i look into it.

hope366:
Thank you very much.
Looking forward to great improvements😄.

hope366:
I've confirmed that it has been fixed.
It's been a wonderful job!

kaorahi:
thx for merging xiaoyifang#5 :)

PR #854, opened 2021-01-30
Fix negative alpha value by too many ctrl+shift+pgdown

kaorahi:
ref. #290


hope366:
Thank you for creating the PR.
I tried this PR, but in my environment, even if I execute Ctrl + Shift + PageDown (Up), there is no change ...

hope366:
I had imagined that a single operation would make a clear change, so it seemed like it wasn't working.
As you say, I was able to see the effect by hitting PageDown / Up many times 😄 (faster if you hold it down)

Issue #852, opened 2021-01-29
About key operation


hope366:
There are some unclear points regarding key controls.
* "ctrl = undo / redo 10 moves" ... Nothing happens when I press "Ctrl".Actually, it works with "ctrl + downArrow" and "ctrl + upArrow", so it is a mistake in the key operation explanation column of pressing and holding the x key.
* "D = show / hide dynamic komi" ... Nothing happens when you press "d".
* "R = replay branch" ... I don't know how to use it.
* ctrl + shift + pageDown: decrease opacity of suggestions ... No particular change is seen.
* DisplayStrings.properties says "key E = e | toggle evaluation coloring", but the e key shows / hides the gtp console.
Should I modify it to "keyE = e | toggle GTP console" and add "keyK = k | toggle evaluation coloring"?

kaorahi:
* ctrl: just as you said :)
* d: broken (#855). See also #832. This key works differently on a forked Leela Zero (#329) and KataGo. No effect on the original Leela Zero.
* r: #436
* ctrl + shift + pageDown: #290. You may need to hit it repeatedly until you notice its effect. (Some keyboards do not recognize this key combination. See also #854.)
* e, k: #836

hope366:
https://github.com/alreadydone/lz/releases
I downloaded clear_nncache.exe and komi + half-0829-3.exe from here and tried it with lizzie-0.7.4.
It worked, but I couldn't confirm the function of Dynamic komi ...
I will not pursue this matter too deeply.

I finally got to know about the replay branch.
From the word "branch", I focused only on the variation tree, but this was a big mistake.
By pressing the "R" key with the mouse cursor placed on the candidate hand on the main board, the change diagram was automatically played 😄
It is pondering off during autoplay, but automatically switches to pondering on when autoplay ends.
I think that this function is often used when you want to check the change diagram slowly, so why not turn it off when autoplay is finished?

kaorahi:
Dynamic komi (D key, #329) is one of unfinished experiments for handicap games before KataGo. You will not need it now since KataGo plays handicap games nicely.

hope366:
Yes, I think KataGo's handicap game is clearly better. However, the research on dynamic comics seems so enthusiastic that I want to realize how effective it is.
I also tried the lizzie-next branch created by @alreadydone san, but it doesn't seem to work ...

PR #857, opened 2021-01-31
Keep pondering on/off after animation by "r" key

kaorahi:
https://github.com/featurecat/lizzie/issues/852#issuecomment-770226346


hope366:
It's working fine.
Thank you.

PR #861, opened 2021-02-03
Fix #859 (overclipping in variation tree)

### LINKED ISSUE #859: The top of the variation tree appears to be cut off
If you move the left figure two steps, it will look like the center figure.
It looks like the main branch and the sub branch are separate.
This is not a bug because it is such a specification, but it may look better if the line is attached to the top like MultiGo (right figure).

kaorahi:


hope366:
Thank you for incorporating the proposal.

Issue #621, opened 2019-10-01
Feature Request: Stop Engine from Running After Auto Analysis


bdrose:
It would be nice to have the engine stop automatically after running Auto Analyze (a). It is pointlessly utilizing the CPU/GPU by doing an insane amount of analysis on the last move if you do not manually stop it.

featurecat:
you can enable it in the configuration. Max analyze time minutes. in the new version I set it to 99999, but you can change it.

featurecat:
ohhh i see what you mean now. yeah that's a good idea, I can't believe I didn't think of that until now haha. I'll implement that for next version, hopefully we can release that much sooner this time.

featurecat:
Fixed with 62d92669b5bbb4c7ddceb4e96abf29f11198b672

PR #863, opened 2021-02-05
not turn off ponder when finished fast analysis.(invoked by keymap a)

xiaoyifang:
after analysis the loaded sgf(invoked by key 'a') ,not toggle ponder status .
keep its original status.


kaorahi:
See #621.

xiaoyifang:
> See #621.

got it ,

I think there are situations when the user (just like me ),want the engine continue to analysis.  users can disable the analyis  himself.

the fast analysis can give a roughly analysis ,while the continued analysis continue to refine the current node  result ,until the users turn it off intentionally.

i'll close the PR anyway.

PR #864, opened 2021-02-05
Fix #862 (insufficient x-shift in variation tree)

### LINKED ISSUE #862: The rightmost variation is not visible in the variation tree
In this SGF, the rightmost variation is not visible in the variation tree even if we select that variation on the board. (Lizzie 0.7.4 on a narrow window.)

~~~
(;(;B[aa](;W[da])(;W[ea](;B[fa])(;B[ga])(;B[ha])(;B[ia])(;B[ja])(;B[ka])(;B[la])(;B[ma])(;B[na])(;B[oa])(;B[pa])(;B[qa])(;B[ra])(;B[sa])))(;B[ab];W[bb]))
~~~

This is a simpler version. getCurLane() returns 2 wrongly for the rightmost variation. It should be 3 correctly.

~~~
(;(;B[aa](;W[da])(;W[ea](;B[fa])(;B[ga])))(;B[ab];W[bb]))
~~~

kaorahi:
I gave up debugging of getCurLane() and use a dummy drawing to get the current lane to fix #862 certainly.

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/VariationTree.java
+++ b/src/main/java/featurecat/lizzie/gui/VariationTree.java
@@ -26,6 +26,7 @@ public class VariationTree {
   private BoardHistoryNode curMove;
   private Rectangle area;
   private Point clickPoint;
+  private int curMoveLane = 0;
 
   public VariationTree() {
     laneUsageList = new ArrayList<Integer>();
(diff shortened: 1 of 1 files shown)


hope366:
What does "x-shift" in the title mean?

kaorahi:
I meant "horizontal offset".

hope366:
I see, thanks.

PR #865, opened 2021-02-05
Accept two or more spaces like "-w  network.gz" in engine commands

kaorahi:


hope366:
With the existing lizzie, the nickname or weight name will be displayed in the engine column by leaving a space between "-w" and "network.gz". If there is no space or if two or more spaces are inserted, the engine will start but the engine field will be blank.
However, by applying this PR, it is possible to prevent the engine column from becoming blank even if there is no space or two or more spaces are inserted.
I interpret it this way, but is this correct?

I haven't felt any inconvenience without this PR, but what motivated you to create this PR?

PR #866, opened 2021-02-07
Fix issues in #556 (variation features)

kaorahi:
in #556:

1. The first hit of UpArrow key was ignored.
2. Stone colors were wrong for the branch that is added by right click of a suggested move on the empty board.

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/BoardPane.java
+++ b/src/main/java/featurecat/lizzie/gui/BoardPane.java
@@ -456,10 +456,9 @@ public class BoardPane extends LizziePane {
     } else {
       if (boardRenderer.isShowingNormalBoard()) {
         setDisplayedBranchLength(boardRenderer.getReplayBranch());
-      } else {
-        if (boardRenderer.getDisplayedBranchLength() > 1) {
-          boardRenderer.incrementDisplayedBranchLength(-1);
-        }
+      }
+      if (boardRenderer.getDisplayedBranchLength() > 1) {
+        boardRenderer.incrementDisplayedBranchLength(-1);
       }
     }
   }
--- a/src/main/java/featurecat/lizzie/gui/BoardPane.java
+++ b/src/main/java/featurecat/lizzie/gui/BoardPane.java
@@ -445,22 +445,7 @@ public class BoardPane extends LizziePane {
   }
 
   public void doBranch(int moveTo) {
-    if (moveTo > 0) {
-      if (boardRenderer.isShowingNormalBoard()) {
-        setDisplayedBranchLength(2);
-      } else {
-        if (boardRenderer.getReplayBranch() > boardRenderer.getDisplayedBranchLength()) {
-          boardRenderer.incrementDisplayedBranchLength(1);
-        }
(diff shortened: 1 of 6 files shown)


kaorahi:
Imported #556 into normal UI.

ref. https://github.com/featurecat/lizzie/pull/867#issuecomment-774804632

hope366:
I tried d0f9d18 and 9886724.
9886724 worked, but d0f9d18 didn't.
Just in case I downloaded https://github.com/kaorahi/lizzie/tree/fix556_var and tried it, but Add As Branch didn't work in normal mode.

kaorahi:
Would you read this essay? It explains what we need. :)

How to Report Bugs Effectively
https://www.chiark.greenend.org.uk/~sgtatham/bugs.html

(Japanese translation)
https://www.chiark.greenend.org.uk/~sgtatham/bugs-jp.html

hope366:
We apologize for the inconvenience.
My report was just "not working". This was not enough.
First of all, regarding this PR, I verified it by overwriting Patched_0.7.4. I overwrote both d0f9d18 and 9886724 and verified. Not verified separately.
Regarding d0f9d18, in normal mode, right-clicking on the candidate hand with no stones on the board did not respond and nothing happened.
If you do the same operation with stones on the board, it will have the effect of "undoing".
In other words, it is exactly the same as before applying d0f9d18.
There is no additional information in the GTP console, and right-clicking on it doesn't do anything.

I haven't told you so far, so I'll give you the following information as well.
Windows 10 Pro 2004
For java, jdk-14.0.1 and jre1.8.0_271 are installed.
The compiler uses apache-maven-3.6.3-bin.

kaorahi:
The word "undo" in your second report may be a key hint. :)

I guess you have set "Avoid Keep Variations" to 0. It disables right-click menu and performs "undo" instead if you have applied 29507eb5 and 9cd6baf in #773.

In the next bug report, it is better to try the default config.txt. Namely, move your config.txt somewhere temporally and start Lizzie without config.txt. Then Lizzie generates the default config.txt.

(This may also solve #868.)

hope366:
For #868, I solved it by using the default config.txt. It was because "Avoid Keep Variations" was set to 0.Thank you for solving the problem.

However, the problem that "Add as branch" cannot be used in normal mode is still unsolved. Even if "Avoid Keep Variations" was the default "30", it didn't work. I tried applying d0f9d18 alone and the result was the same.

The engine used for verification is katago-v1.8.0,
The weight file is kata1-s609.
The configuration file of katago is default_gtp.cfg, and I have changed some settings and used it, but I think that it has nothing to do with this lizzie case.

kaorahi:
Ah, we should've confirmed the engines. I used Leela Zero for quick tests, while you used KataGo.

It was disabled for KataGo. thx for the heads-up!

hope366:
I see, was that so!
It was my lack of consideration that I did not specify the engine used.
I learned how important it is to convey more information when reporting a bug. This time I had a good experience with your help. thank you very much.
I tried 25c878b. It's working fine 😄

PR #867, opened 2021-02-07
Revive stoneless movenums in #223, that was disabled by #436 and #586

kaorahi:
Usage:

1. Show a suggested variation by mouse hover as usual.
2. Push "z" key (keep holding down) to erase the variation stones and show only their move numbers.
3. Push "up/down arrow" keys to show the variation stones step by step.

Push "up arrow" key just after 2. to erase everything.

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/BoardRenderer.java
+++ b/src/main/java/featurecat/lizzie/gui/BoardRenderer.java
@@ -627,7 +627,7 @@ public class BoardRenderer {
         variation = suggestedMove.get().variation;
       }
     }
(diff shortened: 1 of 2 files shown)


hope366:
In normal mode, you can undo or replay a move with the up and down arrow keys or by scrolling up and down with the mouse, regardless of whether the mouse cursor is over the candidate move or not.
In Panel UI mode, on the other hand, if the mouse cursor is over a candidate move, you can check the variation diagram move by move, and if the mouse cursor is somewhere other than the candidate move, you can undo or replay the stones.
I think it would be easier to understand if the normal mode worked the same way as the panel UI mode, what do you think?

PR #873, opened 2021-02-12
Show "pass" mark on the board even when show-move-number is true

kaorahi:
I expect the large "pass" mark on the board even when I select "All" or "Last" for "Show Move Number" in Settings. How do you think?


hope366:
I think this is a good idea, but by applying this PR, the pass mark is no longer displayed when passing with the move number set to hidden.
I don't know if it was a deliberate change, but I think it's better to show the pass mark even when the move number is hidden.

hope366:
Thank you for your prompt response.
I have confirmed that the 92e298d is working fine.

PR #874, opened 2021-02-12
Fix incorrect thick line in variation tree at the first move

kaorahi:


hope366:
We were not able to confirm what changes occurred before and after the application of this PR.
I am observing the area shown by the screenshot below, is this correct?
Is the change so subtle that it is difficult to see with the eyes?
無題

kaorahi:
Would you try this SGF on the default config.txt (normal UI) and compare its result with your previous screenshot?

~~~
(;B[aa])
~~~

hope366:
I tried `(;B[aa])` with normal UI + default config.
The line from the first move to the second move was thicker than usual. This thickness seems to depend on the value of "Winrate Stroke Width".
 I had this setting set to "1" so I couldn't seem to find any difference.
The method of verification was insufficient. I would like to keep in mind to use the default config when checking for bugs.

hope366:
I tried a little more verification.
If you create a branch like the one on the left, it will be a dotted line, but this also happened even if you set the Winrate Stroke Width to "1".
Next, look at the figure on the right.
In this way, when the movement of the main tree is fixed at "1", the subbranch line seems to be a "thick wrong line" forever.(Only if the current move exists in the root node or in the "1" move of the main tree.)
This PR seems to solve these as well.

無題1   無題

PR #890, opened 2021-07-20
Fix #889 (wrong color to play)

### LINKED ISSUE #889: Analytical anomaly with SGF with fixed stone placement
I will explain how to reproduce this bug.
1. Register multiple engines in lizzie in advance.For example, katago-eigen, katago-eigenavx2, katago-opencl, leelazero, etc. You may register multiple items with exactly the same contents.
1. Prepare an SGF with a fixed stone arrangement instead of a normal SGF. Please use the following SGF this time. 
(; AP [MultiGo: 4.2.4] SZ [19] MULTIGOGM [1]
AB [dd] [qd] AW [dp] [pp])
1. After starting lizzie, load the above SGF and analyze it. You should probably see a number around 50%. 
1. Then switch to another engine. At this time, a value exceeding 90% should be displayed. (It should be around 50% originally) 

Also, this bug may or may not occur due to subtle changes in the placement of the stones. For example, the SGF below did not have this bug.
(; AP [MultiGo: 4.2.4] SZ [19] MULTIGOGM [1]
AB [dd] [pd] AW [dp] [pp]) 

This bug occurred in all three types of lizzie: lizzie0.7.3, Lizzie-patched-0.7.4_210704b, Lizzie-improved-v2.7.
The engines I used for verification are katago1.9.1eigen,katago1.9.1eigenavx2, katago1.9.1opencl, leelazero0.17. We were able to confirm this bug in all these engines.  
We have only verified it in a Windows 10 environment.

kaorahi:
(Caution: You also need to apply #766 to handle initial stones correctly in Lizzie 0.7.4.)

This patch fixes both #889 and another issue (handicap stones are not cleared for background engines) in Lizzie 0.7.3:

1. Start Lizzie.
2. Paste `(;AB[dd])`.
3. Switch the engine to Engine1.
4. Click "ClearBoard" button in the toolbar.
5. Switch the engine to Engine0 again.
6. Click "Analyze > Clear analysis" in the menu.

The board is empty at 6, but no candidate appears around the top left corner.

## Notes

This patch is just a workaround. It will be better to specify `player` explicitly in `lz-analyze` like `lz-analyze b 10` instead of `lz-analyze 10`. I did not do it this time because I am not sure whether `Lizzie.board.getData().blackToPlay` (or `Lizzie.board.getHistory().isBlacksTurn()`) is always updated correctly before pondering.

### DIFF
--- a/src/main/java/featurecat/lizzie/rules/Board.java
+++ b/src/main/java/featurecat/lizzie/rules/Board.java
@@ -823,15 +823,10 @@ public class Board implements LeelazListener {
 
   /** Restore move number by node */
   public void restoreMoveNumber(BoardHistoryNode node) {
-    Stone[] stones = history.getStones();
-    for (int i = 0; i < stones.length; i++) {
-      Stone stone = stones[i];
-      if (stone.isBlack() || stone.isWhite()) {
-        int y = i % Board.boardWidth;
-        int x = (i - y) / Board.boardHeight;
-        Lizzie.leelaz.playMove(stone, convertCoordinatesToName(x, y));
-      }
-    }
+    Lizzie.leelaz.clear();
+    boolean blackToPlay = node.getData().blackToPlay;
+    restoreInitialStones(blackToPlay);
+    restoreInitialStones(!blackToPlay);
     int moveNumber = node.getData().moveNumber;
     if (moveNumber > 0) {
       if (node.isMainTrunk()) {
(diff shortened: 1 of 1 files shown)


kaorahi:
Ideally, we should support [PL](https://www.red-bean.com/sgf/properties.html#PL) property in SGF to specify who to play if there are both black and white stones in the initial setup. See also Katrain's smarter way. https://github.com/sanderland/katrain/issues/293

Issue #892, opened 2021-08-26
Analysis


bohemianatlarge:
Maybe this is an issue or maybe not. How can I save the file to sgf without all of the analysis added to the file, as this can make for very large cumbersome sgf files when there are a lot of branches to the trees that I am working on. I just want the moves and their variations I don't want all of the lines and lines of analysis. Thanks

kaorahi:
Do you mean Settings > View > Hold Best Moves To Sgf ?

bohemianatlarge:
I am not sure I have not tried that. I just want to be able to save what I am working on to an SGF file without it adding all that extra information to the file. If that does the trick than thanks!

Unticking the box and saving the file eliminates all the extra info being saved to the file! Thanks mate!

PR #893, opened 2021-08-26
Add tooltips for truncated long labels in config dialog

kaorahi:
workaround for unreadable labels, e.g. #892

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/ConfigDialog.java
+++ b/src/main/java/featurecat/lizzie/gui/ConfigDialog.java
@@ -82,6 +82,7 @@ import javax.swing.LayoutStyle.ComponentPlacement;
 import javax.swing.ListCellRenderer;
 import javax.swing.SpinnerNumberModel;
 import javax.swing.SwingConstants;
+import javax.swing.SwingUtilities;
 import javax.swing.SwingWorker;
 import javax.swing.border.Border;
 import javax.swing.event.ChangeEvent;
@@ -113,9 +114,9 @@ public class ConfigDialog extends JDialog {
   private List<String> fontList;
   private Theme theme;
 
-  public JPanel uiTab;
-  public JPanel themeTab;
-  public JPanel aboutTab;
+  public PanelWithToolTip uiTab;
+  public PanelWithToolTip themeTab;
+  public PanelWithToolTip aboutTab;
   public JButton okButton;
 
   // Engine Tab
@@ -256,7 +257,7 @@ public class ConfigDialog extends JDialog {
     tabbedPane = new JTabbedPane(JTabbedPane.TOP);
     getContentPane().add(tabbedPane, BorderLayout.CENTER);
 
-    JPanel engineTab = new JPanel();
+    PanelWithToolTip engineTab = new PanelWithToolTip();
     tabbedPane.addTab(resourceBundle.getString("LizzieConfig.title.engine"), null, engineTab, null);
     engineTab.setLayout(null);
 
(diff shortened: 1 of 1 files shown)


hope366:
I wasn't sure exactly what kind of change was caused by this PR.
I'm sorry, can you tell me?

kaorahi:
In my Linux environment, Settings > View > Hold Best Moves To Sgf is shown as "Hold Best Moves To ..." actually. This patch shows tooltips by mouse hover on such truncated labels. (maybe needless for Windows)

hope366:
Thank you for letting me know.
In the Windows environment, "Settings"-> "View"-> "Color by Winrate Instead of Visits" is interrupted, which is convenient for displaying here.
The same part is cut off in the Japanese version, but in my "lizzie improved version", the description was changed a long time ago so that all of them are displayed.
This PR made it possible to revert to the previous description 😄

hope366:
It's nice to see the popup display forever while you hover your mouse over it, but is it possible to change that?
If possible, I would be grateful if you could change the engine pull-down menu on the top toolbar as well.

kaorahi:
See kaorahi@6b9e5a44.

hope366:
thank you very much.
Work is very fast as usual 😄

kaorahi:
Additional feature: You can specify different tooltips as the texts after `\n`.

(example 1)

in `src/main/resources/l10n/DisplayStrings.properties`

~~~
LizzieConfig.title.showLcbWinrate=Display Winrate As\nNever use LCB!!
~~~

(example 2: decoration)

~~~
LizzieConfig.title.showLcbWinrate=Display Winrate As\n<html><span style="font-size: 500%; color: red;">Never use LCB!!</span><br><span>It is completely useless for normal users.</span></html>
~~~

hope366:
Is there a way to display multiple lines if the description is quite long?
Also, is it possible to adjust the time from when the mouse cursor is placed until the popup is displayed?

kaorahi:
See the above example 2 for multi-line tooltips. Add the following line into kaorahi@6b9e5a44 to show the tooltips immediately.

~~~
    ToolTipManager.sharedInstance().setInitialDelay(0);
~~~

hope366:
I'm sorry.
I asked a question without checking your post.
Thank you for telling me a convenient method.

hope366:
I was able to confirm that the popup was displayed quickly.

Just adding 6e4eeaf doesn't make the long description multiple lines, is there anything else I need to do?

hope366:
6e4eeaf is a code to display the sentence after "\ n" as a tooltip, isn't it?
Apart from that, is it possible to have a very long description displayed on multiple lines? 

It seems that it is possible to make multiple lines by writing as in Example 2 above, but with this method, it seems that it is necessary to specify for each description.
If this is the case, the amount of work will increase because all items need to be corrected.
If the text is long to some extent, is there a way to automatically fold it and display it in multiple lines?

kaorahi:
Hmm, I don't like to pay too much efforts for it...

(example 3: word wrap)

~~~
LizzieConfig.title.showLcbWinrate=Display Winrate As\n<html><p style="width: 100px;">foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo</p></html>
~~~

hope366:
I'm sorry to have requested many things.
You taught me more than enough.
I will try to make full use of the information given.

kaorahi:
Have you tried the above example 3? It is an example of automatic line breaks in the given width. Or, you can simply insert `<br>` as follows if you don't need decorations.

(example 4: manual line breaks without decoration)

~~~
LizzieConfig.title.showLcbWinrate=Display Winrate As\n<html>abcdefg<br>hijklmn<br>opqrstu<br>vwxyz</html>
~~~

hope366:
Thank you for the additional information.
I didn't know anything about html, but looking at Example 2, I guessed that inserting </ span> <br> <span> would break the line, and I did this and it worked.
It was good to just <br>
I also tried Example 3, but I thought it would be easier to read a line break at a well-separated part, so I adopted the method of manually breaking a line.

PR #896, opened 2021-09-01
Fix: update pondering message after max-analyze-time-minutes

kaorahi:
"pondering on" message was not updated after max-analyze-time-minutes. thx > @hebaeba.

Though I also wrote the following dialog, it is not included in this PR because I am afraid that such warnings may be annoying. How do you think?

~~~
Paused by the time limit (XX min from yyyy-MM-dd HH:mm).

Push spacebar to resume.
Click "Settings > Engine" menu to change "Max Analyze Time".
~~~

(d9f755b5 in https://github.com/kaorahi/lizzie/tree/fix_max_time2)


hope366:
I think it's a good change to display the d9f755b dialog message because it's easier to visually confirm that the analysis is complete.
Is it possible to display this dialog message in multiple languages?

kaorahi:
See fdfa0af0.

hope366:
thank you very much.
As a Japanese version, I tried to create something like the following. 
bunnseki

PR #897, opened 2021-09-03
Show change of winrate bar

kaorahi:
This patch imports one of my favorite features from LizGoban. The change of the winrate bar is shown as the green (= good move) or red (= bad move) rectangle.

Are there more suitable colors for Lizzie?

wrbar_diff

### DIFF
--- a/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
+++ b/src/main/java/featurecat/lizzie/gui/LizzieFrame.java
@@ -1030,6 +1030,20 @@ public class LizzieFrame extends MainFrame {
       g.setColor(Color.BLACK);
       g.fillRect(barPosxB, barPosY, barWidthB, barHeight);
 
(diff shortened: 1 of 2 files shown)


hope366:
I think it's an interesting feature.
The win rate bar of lizzie will be displayed when Nihon Ki-in broadcasts live on youtube, but I think that viewers can enjoy it more by using the version with this function. 

I would just like to mention one question that I have.
The image you attached shows a red rectangle to indicate how bad a move White made, but because it is on top of the black bar, it seems to intuitively recognize that Black made a bad move.

kaorahi:
better ideas are welcome :)

kaorahi:
As the above design seems to confuse some users, I modified it so that Lizzie shows just the previous winrate with a simple gray line. I expect this will not disturb your eyes.
wrbar_diff2

PR #904, opened 2021-11-06
Auto-continuation of ladders

kaorahi:
Play at least 5 ladder moves successively and click the menu Game > Continue ladder. Then the rest of the ladder is played automatically.

This is a cheaper version of lizgoban's [automatic ladder continuation](https://github.com/kaorahi/lizgoban/issues/63). It may be useful for some youtubers in their game reviews. @yzyray, I also expect you will export this to [LizzieYzy](https://github.com/yzyray/lizzieyzy) if you like.

The implementation is very simple. Instead of serious ladder detection, it just detects "periodically repeated patterns" like https://github.com/leela-zero/leela-zero/pull/1941 . So we need at least 5 moves beforehand.


yzyray:
Useful function~thank for sharing,will add in lizzieyzy.

